PR #22, opened 2026-02-11
fix: quality audit - 8 bug fixes + 65 handler integration tests

gourmetfire:
### Bug Fixes

| # | Category | Fix |
|---|----------|-----|
| 1 | **CI** | Enable integration tests (removed `--lib` from `cargo test` in CI) |
| 2 | **Timestamp collisions** | `timestamp_nanos_opt().unwrap_or(0)` → `timestamp_micros()` in `state.rs` (4 sites) + `learning_history.rs` (6 sites). Zero-padded audit keys (`:020`) for correct RocksDB lexicographic ordering |
| 3 | **Auth gate** | Moved `/api/graph/data/{user_id}` from public to protected routes (was accidentally unauthenticated) |
| 4 | **Entity cap** | Sort entities by salience descending, truncate to 10. Truncate context to 150 chars to prevent unbounded growth |
| 5 | **`set_var` UB** | Added `pre_init_ort_runtime()` called before any async work in `main.rs` — eliminates undefined behavior from `std::env::set_var` in multi-threaded context |
| 6 | **ONNX threads** | Added `with_inter_threads(1)` in both `minilm.rs` and `ner.rs`. Platform-specific inter-thread defaults (1 on macOS ARM64, 2 elsewhere) to prevent ONNX thread pool over-subscription |

### Tests Added

- **65 handler integration smoke tests** covering all 19 handler groups
- Auth middleware tests verify public routes are open and protected routes reject missing API keys
- End-to-end remember→recall cycle test
- Test harness mirrors `main.rs` auth middleware layering for accurate API coverage
- **All 424 lib tests + 65 integration tests pass**

### Files Changed (11)

- `.github/workflows/ci.yml` — CI fix
- `Cargo.toml` — added `http-body-util` dev-dep
- `src/embeddings/minilm.rs` — ONNX runtime init + thread config
- `src/embeddings/ner.rs` — ONNX thread config
- `src/handlers/mod.rs` — test_helpers module declaration
- `src/handlers/router.rs` — auth gate fix
- `src/handlers/state.rs` — timestamp + entity cap + context truncation
- `src/handlers/test_helpers.rs` — NEW: shared test utilities
- `src/main.rs` — pre_init_ort_runtime call
- `src/memory/learning_history.rs` — timestamp fixes
- `tests/handler_tests.rs` — NEW: 65 integration tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -224,6 +224,10 @@ harness = false
 name = "hebbian_benchmarks"
 harness = false
 
+[[bench]]
+name = "softmax_benchmarks"
+harness = false
+
 [profile.release]
 opt-level = 3
 lto = true
--- a/benches/softmax_benchmarks.rs
+++ b/benches/softmax_benchmarks.rs
@@ -1,0 +1,59 @@
+//! Micro-benchmarks for argmax_softmax vs allocating softmax + max_by.
+//!
+//! Validates the perf claim that fused argmax_softmax avoids a per-token Vec
+//! allocation on the NER hot path.
+
+use criterion::{black_box, criterion_group, criterion_main, BenchmarkId, Criterion};
+use shodh_memory::embeddings::ner::argmax_softmax;
+
+/// Baseline: allocating softmax then argmax (the old code path).
+fn softmax_then_argmax(logits: &[f32]) -> Option<(usize, f32)> {
+    if logits.is_empty() {
+        return None;
+    }
+    let max_logit = logits.iter().cloned().fold(f32::NEG_INFINITY, f32::max);
+    let exp_sum: f32 = logits.iter().map(|x| (x - max_logit).exp()).sum();
+    let probs: Vec<f32> = logits.iter().map(|x| (x - max_logit).exp() / exp_sum).collect();
+    probs
+        .iter()
+        .enumerate()
+        .max_by(|a, b| a.1.partial_cmp(b.1).unwrap_or(std::cmp::Ordering::Equal))
+        .map(|(idx, &prob)| (idx, prob))
+}
+
+fn bench_argmax_softmax(c: &mut Criterion) {
+    let mut group = c.benchmark_group("argmax_softmax");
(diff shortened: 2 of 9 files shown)


varun29ankuS:
Hey @gourmetfire — resolved everything in #23.

**What made it in (all the good stuff):**
- Auth gate fix, DashMap TOCTOU, WriteBatch atomicity, sync main — all applied
- argmax_softmax fusion, entity take(), streaming rotation, context pre-compute
- All 65 handler integration tests + TestHarness infrastructure
- Criterion benchmark, entity cap config, NER timeout metric
- CI now runs integration tests

**What I dropped (overlap with main or data-safety):**
- `minilm.rs` — main already has a better fix for the hash spreading (`wrapping_mul(7)` + fresh hashers from adc4764). Your version would have regressed that.
- NER lock timeout refactor — already landed in 977be37 with the same pattern.
- `:020` audit key padding — this would silently break lexicographic ordering on existing RocksDB data. Existing installs have unpadded keys, and mixed formats would corrupt range scans and rotation. Kept the current format.

**What I fixed during merge:**
- `argmax_softmax` needed `pub` not `pub(crate)` — benchmarks in `benches/` are external crates, so `pub(crate)` won't compile.
- First apply had `timestamp_micros()` where the system expects nanos — caught and reverted.
- `prune_old_events` had a data-loss path in the initial apply — fixed in follow-up commit.

**Feedback:**
The TOCTOU fix, WriteBatch atomicity, and sync main are really solid catches. Those are the kind of subtle correctness issues that most audits miss. The 65 integration tests are a huge addition — we went from zero handler test coverage to comprehensive.

The issues that came up (key padding, minilm regression, NER overlap) are all integration-level gaps from working 7 commits behind main. Nothing wrong with the approach, just stale context. Now that this is merged, future PRs against current main should be smoother.

Once #23 merges, I'll close this one with a cross-reference. Thanks for the contribution.

PR #134, opened 2026-03-26
fix(tui): improve detail panel rendering, text wrapping, and SSE context

phreakocious:
TUI & Rendering Fixes
   - UTF-8 Word Wrapping: Fixed a bug where multi-byte characters were causing word-wrapping to miscount line lengths, which would break the UI layout. We now use char_indices() to correctly split strings at valid byte boundaries.
   - Uncapped Content: Removed the arbitrary 6-line hard cap on content in the detail panel. The text will now fill all available vertical space.
   - Better Layout: Moved Tags to appear above the main content block so they are always visible and don't get pushed off-screen by long memories.
   - Dynamic Sizing: Increased the default height of the detail panel from a fixed 12 rows to dynamically take up half the screen (or a minimum of 14 rows) to provide a much better reading experience.

  Backend / SSE Enhancements
   - Expanded Preview: Increased the content_preview length in SSE broadcasts from 100 characters to 500 characters to prevent memories from looking prematurely truncated in the TUI stream.
   - Tag Broadcasting: Added a new entities field to MemoryEvent. The backend now explicitly broadcasts tags during CREATE/UPDATE events so the TUI can display them live in the activity log.

  Chore / Performance
   - Docker Build Cache: Optimized the Dockerfile to cache the dependency build layer separately from the application source code. This will massively speed up Docker rebuilds during development.


varun29ankuS (approved):
Clean PR — all changes look good.

- UTF-8 word wrapping fix with `char_indices()` is the correct approach
- Tags above content is better UX (always visible)
- Dynamic panel sizing is a nice improvement
- SSE content_preview bump to 500 chars and `entities` field on MemoryEvent are thorough — every construction site updated
- Docker cache layer optimization is solid

Thanks for the contribution!

PR #135, opened 2026-03-26
fix: Zenoh auth, recall I/O reduction, env safety hardening

Portll:
Zenoh transport authentication
— Add SHODH_ZENOH_API_KEY shared-secret with constant-time comparison in authenticate_payload()
— Gate every handler entry (remember, recall, forget, stream, mission) on key match
— Skip auth when key is None for backwards-compatible local-only deployments
— Warn on 0.0.0.0 listen endpoints without auth configured
— Fix doc examples from 0.0.0.0 to 127.0.0.1

Recall hot path I/O reduction
— Short-circuit graph density RocksDB prefix scans when onto_intent.confidence < ONTOLOGICAL_MIN_CONFIDENCE
— Scope Layer 4.9 ontological re-ranking to top 2×max_results candidates instead of all fused results (cuts ~500 RocksDB reads to ~20)
— Cache entity labels in HashMap<Uuid, Option<Vec<EntityLabel>>> during spreading activation so get_entity() is called at most once per entity across all hops
— Remove double-traversal fallback that retried traverse_weighted unfiltered when filtered results < 3 — spreading activation penalties via ONTOLOGICAL_RELATION_PENALTY already handle this
— Pass OntologicalIntent from mod.rs into spreading_activation_retrieve_with_stats instead of recomputing infer_ontological_intent twice

env and key resolution safety
— Wrap std::env::set_var calls in server::run() with unsafe blocks and SAFETY invariant comments (no threads before tokio runtime build) — prepares for Rust 2024 edition
— Replace chained || API key resolution in MCP server with explicit if/else that logs source env var name without revealing the key, warns on SHODH_DEV_API_KEY fallback

### DIFF
--- a/mcp-server/index.ts
+++ b/mcp-server/index.ts
@@ -60,14 +60,26 @@ const SANDBOX_MODE = process.env.SMITHERY_SANDBOX === "true";
 //   3. First key from SHODH_API_KEYS (matches server production config)
 //   4. Auto-generate for local servers (passed to server as SHODH_DEV_API_KEY)
 //   5. Error for remote servers
-let API_KEY = process.env.SHODH_API_KEY
-  || process.env.SHODH_DEV_API_KEY
-  || (process.env.SHODH_API_KEYS?.split(",")[0]?.trim())
-  || (SANDBOX_MODE ? "sandbox" : "");
+let API_KEY = "";
+let apiKeySource = "";
+if (process.env.SHODH_API_KEY) {
+  API_KEY = process.env.SHODH_API_KEY;
+  apiKeySource = "SHODH_API_KEY";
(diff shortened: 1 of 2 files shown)


varun29ankuS (approved):
Solid PR — the recall I/O reductions are well thought out. Entity label cache across both spreading activation passes, density scan short-circuit, and scoping re-ranking to 2×max_results should meaningfully reduce RocksDB reads on the hot path. Zenoh auth and env safety changes look clean too.

Two minor things to track separately — `constant_time_eq` should avoid leaking key length via the early-return length check, and the recall queryable should reject payload-less queries when auth is configured. Will create issues for both. Neither should block this merge.

Approving.

PR #232, opened 2026-04-14
feat: add graph export with JSON API, GEXF output, and CLI

phreakocious:
## Summary
- Adds `GET /api/graph/{user_id}/export` returning the full memory graph as JSON, with optional entity/relationship filters
- Adds GEXF output format for import into Gephi and similar graph-viz tools
- Adds `shodh export-graph` CLI command (async reqwest client)
- Adds `GraphMemory::get_all_episodes()` bulk iterator using `fill_cache(false)` so full scans don't evict hot data from the block cache

## Files
- `src/graph_memory.rs` — `get_all_episodes()`
- `src/handlers/export.rs` — new module: node/edge builders, JSON + GEXF serializers, tests
- `src/handlers/{mod,router}.rs` — wiring
- `src/cli.rs` — `export-graph` subcommand

## Test plan
- [x] `cargo check --all-targets` clean
- [x] `cargo clippy --lib` — no new warnings
- [x] `cargo test --lib handlers::export` — 13 tests pass, including an XML-escape regression test
- [x] End-to-end: ran a local server on live graph data, `shodh export-graph <user> --format gexf` produces GEXF that passes `xmllint --noout` and opens cleanly in Gephi; entity nodes and relationship edges render with proper XML escaping (apostrophes, angle brackets, and stripped C0 control characters that XML 1.0 forbids in attribute values)

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -164,6 +164,44 @@ enum Commands {
         args: Vec<String>,
     },
 
+    /// Export knowledge graph as JSON or GEXF
+    ExportGraph {
+        /// User ID whose graph to export
+        user_id: String,
+
+        /// Output format: json or gexf
+        #[arg(long, default_value = "json")]
+        format: String,
+
+        /// Node types to include (comma-separated: entities,memories,episodes)
+        #[arg(long, default_value = "entities,memories,episodes")]
+        include: String,
+
+        /// Minimum importance threshold for memory nodes
+        #[arg(long, default_value_t = 0.0)]
+        min_importance: f32,
+
+        /// Include embedding vectors
+        #[arg(long)]
+        include_embeddings: bool,
+
+        /// Output file (stdout if omitted)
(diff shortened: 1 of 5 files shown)


varun29ankuS (changes requested):
Hi @phreakocious, thanks for the PR. One blocker though.

`get_all_episodes()` in `graph_memory.rs` uses raw `bincode::serde::decode_from_slice` instead of `crate::serialization::try_decode::<EpisodicNode>()`. All episodes are written with postcard (format-tagged), so this will silently return empty results on any current database. See `get_all_entities()` at line 5023 and `get_episode()` at line 2832 for the correct pattern.

Once that's fixed, good to merge.

varun29ankuS (approved):
Thanks @phreakocious — fix looks good, matches the existing decode pattern exactly. Approving and merging. Appreciate the thorough tests too.

Issue #270, opened 2026-05-02
[recall-harness] Per-pipeline-layer attribution mode


varun29ankuS:
## Context

shodh's recall is a 5-layer pipeline (vamana → spreading → BM25/RRF → cross-encoder rerank → fact boost). Without per-layer attribution, a +8 NDCG embedder uplift can be silently destroyed by the rerank step and we'd never know which layer is the culprit.

Blocked by #266.

## Scope

`recall-eval --layer <mode>` runs each query in 6 modes and reports metrics per mode:

| Mode | Stages enabled |
|---|---|
| `vamana_only` | ANN search only |
| `+spreading` | + graph activation |
| `+bm25` | + lexical fusion (RRF) |
| `+rerank` | + cross-encoder |
| `+facts` | + fact-aware boost (Layer 4.8) |
| `full` | everything (production) |

Each mode runs against the same query set, gets independent metrics. Output JSON has per-layer breakdown.

## Acceptance

- `--layer all` runs all 6 modes in one invocation
- Output JSON `layers.{mode}.{metric}` populated for each
- Documentation explains how to interpret deltas
- Known invariant: `full.ndcg@10 >= vamana_only.ndcg@10` (each layer should add value)

PR #278, opened 2026-05-04
feat(recall): RH-9 — grow smoke corpus from 30 to 108 cases (drops noise floor)

varun29ankuS:
First in the **Y-plan** sequence triggered by PR #277's noise-vs-regression failure analysis. Grows the L1 smoke suite from 30 → 108 cases so the gate has the statistical power to distinguish real regressions from floating-point ordering noise.

## The math problem this fixes

With 30 cases, **each case contributes 3.33% to any metric**. A single rank-flip on one query (e.g., a tie-break going the other way on a different CPU) moves `p@1` by 3.33%, which is *larger than the 2% gate tolerance*. Result: PRs with zero source changes can flip the gate red, and real <3% regressions are invisible. The signal-to-noise ratio is structurally below 1.

At 108 cases the per-case quantum drops to **0.93%**, comfortably below the 2% threshold.

| | before | after |
|---|---|---|
| total cases | 30 | 108 |
| per-category | 5 | 18 |
| per-case quantum | 3.33% | **0.93%** |
| corpus items referenced | 44 / 80 (55%) | 80 / 80 (100%) |

## What's in the new cases

- Every previously-uncovered corpus item now has a query pointing at it (36 new single-target cases).
- Multi-relevant queries that span 2–4 items to exercise the fact-aware fusion layer (e.g., "Trace the path from a memory write to the postcard format tag that protects it.").
- Adversarial negation queries against project rules (`Co-Authored-By` footer, direct push to main, `cargo build` in agent sessions) so the gate catches retrieval drift on operational facts, not just technical ones.
- Three temporal cases pinned to 2026-MM dates so a clock-sensitive ranking bug would surface here first.

## Validator changes

```rust
// src/recall_harness/fixtures.rs
pub const CASES_PER_CATEGORY: usize = 18;   // was 5
pub const TOTAL_SMOKE_CASES: usize = CASES_PER_CATEGORY * 6;
```

`validate_smoke_suite` enforces both — the 14 fixture-validator unit tests still pass.

## Y-plan sequence (this PR is step 1 of 5)

1. **RH-9 (this PR)** — grow corpus to 100+ cases ✅
2. **RH-10** — determinism in retrieval (stable tie-break by UUID, `HashMap` → `BTreeMap` audit, ONNX intra-op = 1, `RAYON_NUM_THREADS=1` for harness)
3. **RH-11** — Rig 1: byte-identical-rank-list determinism gate (separate workflow, sub-second runtime, zero tolerance)
4. **RH-12** — `--repeats N` flag in `recall-eval`, median + IQR aggregation, drop `p@1` from gating metrics
5. **RH-13** — re-baseline on the CI runner so dev-laptop drift stops contaminating the comparison

PR #277 (RH-5) is now in **draft**; it re-lands as a thin wrapper after this sequence.

## Out of scope

- Baseline regeneration — separate PR (RH-13). The current `tests/recall/baseline.json` is intentionally left as-is so this PR does not entangle two changes.
- Layer-attribution columns in the diff table — RH-8 (#270).

## Notes for reviewers

The new queries are written to be answerable by a competent engineer reading the corpus, not by any one keyword match. If you find a query that has an obvious shortcut answer (one shared rare term), flag it — the suite is meant to stress-test semantic recall, not lexical lookup.


varun29ankuS:
Landed: the smoke suite is at 108 cases on main with the baseline regenerated. The synthetic smoke corpus is also being retired now that real benchmarks (LoCoMo held-out, LongMemEval-S) are wired in. Closing.

PR #279, opened 2026-05-05
feat(recall): RH-10 + RH-11 — determinism in retrieval + Rig 1 gate

varun29ankuS:
## Summary

Closes the non-determinism in the retrieval pipeline that was making
the L1 smoke suite flap on cross-environment runs (the failure mode
that surfaced on PR #277). Two layers of fix plus a permanent gate.

### RH-10 — eliminate the sources of non-determinism

**Stable tie-breaks** at every sort site that ranks float scores. Where
two candidates have equal scores, \`f32::total_cmp\` leaves their order
to whatever the underlying container last produced — which is
\`HashMap\` iteration order in several places, i.e. random per process.

- Vector index layer (\`vector_db/{vamana,spann,pq}.rs\`): tie-break by
  \`id\` (the only stable key available cheaply at that depth).
- Retrieval layer (\`memory/retrieval.rs\`, \`memory/mod.rs\`,
  \`relevance.rs\`): score → \`created_at desc\` → \`id\`. Two-layer scheme
  because the index doesn't have timestamps but the retrieval layer
  does, and recency is a meaningful tiebreaker for users.

**Thread pinning** in the harness:

- \`SHODH_ONNX_THREADS=1\` — MiniLM/NER intra-op runs single-threaded.
- \`RAYON_NUM_THREADS=1\` — any \`par_iter()\` in scoring runs serially.

Multi-threaded float reductions accumulate sums in non-deterministic
order, which is enough to flip ranks at the fourth-decimal level even
without any source change. Both vars are only set if the caller hasn't
overridden them, so production callers are unaffected.

### RH-11 — Rig 1: determinism gate

\`tests/recall_determinism.rs\` runs the full L1 smoke suite **twice**
in the same process, against fresh storage dirs, and asserts:

1. **Byte-identical rank lists per case.** \`CaseRankList.retrieved\`
   carries **stable corpus item IDs**, not the random per-run UUIDs
   assigned by \`MemorySystem::remember\` — UUIDs would always diverge
   and would mask real determinism wins behind noise.
2. **Bit-exact metric aggregates.** \`f64::to_bits()\` equality on
   ndcg@10, recall@10, precision@10, mrr, p@1, map — including
   per-category aggregates. If rank lists are equal but metrics
   differ, the metric computation has its own non-determinism.

If the gate ever fires, do **not** relax it — diagnose the new source
of non-determinism. RH-5 (CI gate) and RH-13 (CI re-baseline) both
depend on byte-identical runs to be statistically meaningful.

### Local verification

Two consecutive \`recall-eval\` runs against the smoke fixture produced
**byte-identical** report metrics:

\`\`\`
ndcg@10:    0.8231 → 0.8231  (vs baseline 0.8065, +0.0167)
recall@10:  0.8833 → 0.8833  (vs baseline 0.8500, +0.0333)
p@1:        0.8333 → 0.8333  (vs baseline 0.8333, restored from 0.800
                              that was tripping PR #277's gate)
mrr:        0.9000 → 0.9000
map:        0.7653 → 0.7653
\`\`\`

The temporal category lifted +0.10 — a side-effect of the
\`created_at desc\` tiebreaker in the retrieval layer, which favours
recent memories on score ties (correct behaviour for "what did we
decide last week" style queries).

### Why a separate test for breakage fixes

\`fix(tests)\` lands four pre-existing E0063 errors that blocked the
test suite from compiling — \`EntityNode.selectivity\`,
\`RelationshipEdge.{endpoint_selectivity, forman_curvature}\`, and
\`ReinforcementStats.{entity_edges_reinforced, prediction_error_multiplier}\`.
These were added by upstream PRs (#258, #260, #261) but the test
files used struct literals so the additions broke them silently —
\`cargo test\` couldn't compile until each was filled with safe defaults.

## Commits

- \`8d83f86\` fix(vector_db): RH-10 stable tie-break in vamana/spann/pq
- \`ea7f3b8\` fix(retrieval): RH-10 score → recency → id tie-break
- \`c7cbf83\` fix(recall-harness): RH-10 pin ONNX & rayon to 1 thread
- \`1908177\` fix(tests): add missing struct fields after graph + reinforcement upgrades
- \`b97b125\` feat(recall-harness): RH-11 — determinism gate (Rig 1)

## Test plan

- [x] \`cargo check --tests\` clean
- [x] \`cargo clippy\` clean
- [x] \`cargo test --test recall_determinism\` passes locally (background run)
- [x] Two recall-eval runs produce byte-identical metrics
- [ ] CI: all build matrices green
- [ ] CI: \`recall_determinism\` test passes on Linux + macOS
- [ ] After merge: re-trigger PR #277 to confirm gate stops flapping

### DIFF
--- a/src/memory/mod.rs
+++ b/src/memory/mod.rs
@@ -1249,7 +1249,11 @@ impl MemorySystem {
             let temporal_b = Self::calculate_temporal_relevance(age_days_b);
             let score_b = b.importance() * temporal_b;
 
-            score_b.total_cmp(&score_a)
+            // Score desc → recency desc → MemoryId asc for deterministic ordering.
+            score_b
+                .total_cmp(&score_a)
+                .then_with(|| b.created_at.cmp(&a.created_at))
+                .then_with(|| a.id.cmp(&b.id))
         });
 
         memories.truncate(query.max_results);
@@ -2024,7 +2028,8 @@ impl MemorySystem {
                         )
                     })
                     .collect();
-                r.sort_by(|a, b| b.1.total_cmp(&a.1));
+                // Score desc; tie-break by MemoryId asc for deterministic graph candidate order.
+                r.sort_by(|a, b| b.1.total_cmp(&a.1).then_with(|| a.0.cmp(&b.0)));
                 r.truncate(200);
                 if !r.is_empty() {
                     tracing::debug!(
@@ -2557,7 +2562,11 @@ impl MemorySystem {
             // Only look up graph entities for the top 2x max_results candidates,
             // not all fused results (avoids 100s of RocksDB reads).
             let mut res: Vec<_> = fused.into_iter().collect();
-            res.sort_by(|a, b| b.1.total_cmp(&a.1));
+            // Score desc; tie-break by MemoryId for stable rerank-budget cutoff.
+            // The cutoff at `rerank_budget` makes order at the boundary semantically
+            // important — without tie-break, equal-score boundary candidates can swap
+            // in/out of the rerank window across runs.
+            res.sort_by(|a, b| b.1.total_cmp(&a.1).then_with(|| a.0.cmp(&b.0)));
             let rerank_budget = query.max_results * 2;
 
             if use_ontology_rerank && !onto_intent.expected_labels.is_empty() {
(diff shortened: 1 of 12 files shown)


varun29ankuS:
Consolidating onto PR #277 — those commits are the load-bearing fix for #277's L1 Smoke Suite flapping. No reason to keep them in a separate PR.

PR #284, opened 2026-05-20
fix(security): harden webhooks, rate limiting, errors, and HTTP boundaries

Portll:
## Summary

Fixes security, correctness, and resilience findings surfaced by a
`portll-security` walk-tree audit and a `breakers` audit. Every
behavior-changing fix is **secure-by-default with an explicit override flag**,
so no existing deployment silently breaks.

Headline issues: webhooks processed **unsigned payloads** when no
`*_WEBHOOK_SECRET` was configured (unauthenticated memory poisoning), and
public routes had **no rate limiting** (unauthenticated flood vector).

## Override flags (secure default → opt-out)

| Variable | Default | Effect |
|----------|---------|--------|
| `SHODH_ALLOW_UNSIGNED_WEBHOOKS` | `false` | Webhooks are rejected (503) when no secret is set. `true` restores permissive processing. |
| `SHODH_PUBLIC_RATE_LIMIT` | `true` | Public routes are rate-limited. `false` exempts them (probe routes are never limited regardless). |
| `SHODH_METRICS_PUBLIC` | `false` | `/metrics` requires auth. `true` exposes it to unauthenticated scrapers. |
| `SHODH_ENFORCE_HTTPS` | `false` | `true` rejects insecure `http://` `GITHUB_API_URL`/`LINEAR_API_URL` overrides. |

## Fixes

**Security**
- **Webhooks fail closed**: `/webhook/{linear,github}` reject unsigned requests when no secret is configured, instead of processing them with only a warning.
- **Public-route rate limiting**: the governor is applied to public routes (webhooks, context status, graph viewer). Health probe routes are split into a dedicated router that is *never* rate-limited (avoids spurious k8s restarts).
- **`/metrics` authenticated** by default (Prometheus can authenticate via `Authorization: Bearer`).
- **GraphQL injection fixed**: Linear `fetch_issues` passes `team_id`/`updated_after` through GraphQL `variables` instead of string-interpolating them into the query.
- **5xx responses sanitised**: in production, internal errors return a generic message and log full detail server-side; 4xx errors are unchanged.
- **Malformed webhook signatures** return 400, not 500.

**Correctness / resilience**
- Fixed a **UTF-8 panic** in `extract_json` (a char-count was used as a byte offset → mid-codepoint slice).
- HTTP client timeouts (30s) on the GitHub/Linear clients; a 600s overall-timeout backstop on model/runtime downloads.
- MCP server: non-idempotent writes are no longer retried (duplicate-memory risk); the `proactive_context` in-flight guard is reset on error; the server version is read from `package.json`.
- `Stemmer` is cached per-thread instead of rebuilt per call.

**Docs**
- `SECURITY.md` gains a secure-defaults table; the new `SHODH_*` flags are documented in the configuration help output.

## Verification

- `cargo fmt`, `cargo check` — pass.
- `cargo clippy` — changed files are warning-free (pre-existing warnings in untouched files remain).
- Full `cargo test` was **not** run locally (memory-constrained environment) — CI should run it. New unit tests cover `is_insecure_remote_url` and the Linear query/variable construction (including an injection-attempt test).

## Follow-ups (deferred — out of scope here)

- Automatic re-embedding sweep for memories embedded while the embedding circuit breaker is open (feature-sized).
- CORS production deny-by-default (would break existing deployments).
- `GitHubClient`/`LinearClient` token zeroize-on-drop.

## Live-test gate

The Linear `fetch_issues` rewrite uses a parameterized query; the `$filter: IssueFilter` type name should be confirmed against the live Linear API in a sync test. The inner filter field names (`team`/`id`/`eq`/`updatedAt`/`gte`) are unchanged from the prior working query.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3710,9 +3710,9 @@ dependencies = [
 
 [[package]]
 name = "rustls-webpki"
-version = "0.103.10"
+version = "0.103.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "df33b2b81ac578cabaf06b89b0631153a3f416b0a886e8a7a1707fb51abbd1ef"
+checksum = "61c429a8649f110dddef65e2a5ad240f747e85f7758a6bccc7e5777bd33f756e"
 dependencies = [
  "ring",
  "rustls-pki-types",
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -29,3 +29,20 @@ If you discover a security vulnerability in shodh-memory, please report it priva
 - Disclose publicly before a fix is available
 
 We appreciate responsible disclosure.
+
+## Secure Defaults & Hardening
+
+shodh-memory ships secure-by-default. Each behavior below is enforced unless an
+operator explicitly opts out via an environment variable.
+
+| Area | Secure default | Override (use only if you understand the risk) |
+|------|----------------|-------------------------------------------------|
+| **Webhooks** | `/webhook/*` requests are rejected (HTTP 503) unless the matching `LINEAR_WEBHOOK_SECRET` / `GITHUB_WEBHOOK_SECRET` is configured, so every webhook is HMAC-verified. | `SHODH_ALLOW_UNSIGNED_WEBHOOKS=true` processes unsigned webhooks. |
+| **Rate limiting** | Public routes (webhooks, context status, graph viewer) are rate-limited. Health probe routes (`/health*`) are never rate-limited. | `SHODH_PUBLIC_RATE_LIMIT=false` exempts public routes; `SHODH_RATE_LIMIT=0` disables rate limiting entirely. |
+| **Metrics** | `/metrics` requires API-key authentication (`X-API-Key` or `Authorization: Bearer <key>`). | `SHODH_METRICS_PUBLIC=true` exposes `/metrics` without auth. |
+| **Integration API URLs** | An insecure `http://` override of `GITHUB_API_URL` / `LINEAR_API_URL` for a non-localhost host is warned about. | `SHODH_ENFORCE_HTTPS=true` rejects such overrides and uses the secure default. |
+| **Error responses** | In production (`SHODH_ENV=production`) 5xx responses return a generic message; full detail is logged server-side only. | — |
+| **API authentication** | All `/api/*` routes require an API key. Production refuses authenticated requests when no key is configured. | — |
+| **Transport** | Security headers (`X-Frame-Options`, CSP, `X-Content-Type-Options`, and HSTS in production) are always set. TLS is strongly recommended for any non-localhost deployment. | — |
+
+See the configuration documentation for the full list of variables.
--- a/mcp-server/index.ts
+++ b/mcp-server/index.ts
@@ -154,9 +154,36 @@ const MAX_LIMIT = 250;              // Max results per query
 // TOKEN TRACKING - Context window awareness (SHO-115)
 // =============================================================================
 
-// Token budget configuration (default 100k tokens, ~400k chars)
-const TOKEN_BUDGET = parseInt(process.env.SHODH_TOKEN_BUDGET || "100000", 10);
-const ALERT_THRESHOLD = parseFloat(process.env.SHODH_ALERT_THRESHOLD || "0.9");
+// Parse a numeric env var with validation. A malformed (NaN) or out-of-range
+// value would otherwise silently break token tracking — e.g. a NaN budget makes
+// `tokens / budget` NaN so alerts never fire; a 0 budget makes it Infinity so
+// they always fire. Reject anything outside [min, max] and use the default.
+function parseEnvNumber(
+  raw: string | undefined,
+  fallback: number,
+  min: number,
+  max: number,
+): number {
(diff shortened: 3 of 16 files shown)


varun29ankuS (commented):
Thanks John — this is solid work overall and the diff cleanly implements every claim in the description. The `client_message()` sanitisation, the GraphQL `variables` rewrite, the probe-route split, the `extract_json` char-boundary fix and the MCP retry-idempotency change all check out. Two issues I'd like fixed before merge, plus a handful of smaller things that can be follow-ups.

### Must fix before merge

**1. Webhook fail-closed should return 4xx, not 503.**

In `src/handlers/integrations.rs`, both `linear_webhook` and `github_webhook` return `AppError::ServiceUnavailable("Webhook signature verification is not configured on this server")` when a `*_WEBHOOK_SECRET` is unset and `SHODH_ALLOW_UNSIGNED_WEBHOOKS != true`. Two problems:

- Linear and GitHub treat 503 as transient and will retry-storm forever. Operator misconfiguration shouldn't trigger upstream redelivery loops.
- In production, the 5xx body is sanitised by your own `client_message()`, so the client also has no actionable signal. The logs are right; it's the wire response that misclassifies the failure as "try again later".

`501 Not Implemented` (or any 4xx) stops the redelivery loop and lets a 4xx-shaped message survive sanitisation.

**2. Rate-limit buckets are split between public and protected routers.**

In `src/server.rs`, `make_governor_layer` is called twice (once for `public_routes`, once for `protected_routes`). Each call builds a *fresh* `GovernorConfigBuilder::default()...finish()` and a fresh `GovernorLayer::new(governor_conf)`. The two layers have disjoint per-IP state. A single attacker IP can sustain `rate_limit_per_second × 2` against the server because the bucket on `/webhook/*` is independent of the bucket on `/api/*`.

The code comment says "Each call yields independent per-IP rate-limit buckets ... without depending on GovernorLayer being Clone." But `tower_governor` 0.8 (your pinned version) *does* impl `Clone` for `GovernorLayer` — I checked the source at `tower_governor-0.8.0/src/lib.rs`:

```rust
impl<K: KeyExtractor, M: RateLimitingMiddleware<QuantaInstant>, RespBody> Clone
    for GovernorLayer<K, M, RespBody>
{
    fn clone(&self) -> Self {
        Self { config: self.config.clone(), error_handler: self.error_handler.clone() }
    }
}
```

The internal `config` is `Arc<GovernorConfig<...>>` and `GovernorLayer::new` accepts `impl Into<Arc<GovernorConfig<K,M>>>`. So either share an `Arc<GovernorConfig>` between two `GovernorLayer::new(arc.clone())` calls, or build one `GovernorLayer` and `.clone()` it. Either path produces shared per-IP state.

At minimum, please document in the operator-facing config help that `rps` is per-router so a "4000 rps" setting permits 8000 sustained from one IP.

### Minor (follow-up tickets OK)

- `is_insecure_remote_url` in `src/integrations/mod.rs` is hand-rolled and misclassifies (a) URLs with userinfo — `http://user:pass@localhost/` is correctly localhost but the parser sees the `:` in `user:pass` first and treats `"user"` as the host; (b) IPv6 bracketed hosts — `http://[::1]:8080/` doesn't match the `::1` loopback test because the split keeps the leading `[`. `url::Url` is already in your transitive deps. For a function gating a security decision it's worth using a real parser.
- `mcp-server/index.ts` `apiCall`: `method === "GET"` is case-sensitive. Every call site uses upper today, but `method.toUpperCase() === "GET"` removes the footgun. Also worth a one-liner doc that `HEAD`, `OPTIONS`, `PUT` are now non-retryable too (strictly correct, slightly stricter than RFC 7231's idempotency definition).
- `src/integrations/github.rs` `GitHubClient::new` and `src/integrations/linear.rs` `LinearClient::new` use `.expect("Failed to build … HTTP client")` on `reqwest::ClientBuilder::build()`. That panics the whole server on a transient TLS-init / system-cert-store failure. Propagate as `Result<Self>`.
- `LINEAR_ISSUES_QUERY`: `$filter: IssueFilter` is probably the right type name for Linear's current schema, but a sync-test gated by a real-Linear env var would catch a future Linear rename. Not blocking.
- Precedence: `SHODH_RATE_LIMIT=0` silently overrides `SHODH_PUBLIC_RATE_LIMIT=true`. Correct hierarchy, but worth one sentence in the operator docs so people don't wonder why their public limiter "isn't working".

### Confirming intentional

- Probe-route split (`build_probe_routes` merged without the governor layer): correct — k8s liveness/readiness probes won't trip rate limits.
- The MCP retry-idempotency change disables retry on writes even for HTTP-idempotent methods (`PUT`, `DELETE`). Conservative but right — preserves write-once semantics when the API may have committed before ACK.
- GraphQL injection fix: `LINEAR_ISSUES_QUERY` is a `const &str` with no interpolation; `team_id`/`updated_after` flow through `build_issues_variables` into the `variables` JSON object. The injection-attempt test correctly verifies the escape boundary.
- 5xx sanitisation: full error logged via `tracing::error!` in `IntoResponse for AppError` *before* body sanitisation. Server-side debug capability preserved.
- Zenoh auto-topic security warning: accurate — auto-topics are unauthenticated by design and `authenticate_payload` doesn't apply to them.

Happy to land this immediately after (1) and (2) are addressed.

Issue #10, opened 2026-01-18
Documentation mismatch: rate limit defaults


KTamayo:
## Summary

Code comments, help text, and `.env.example` show different default values than the actual implementation.

## Environment

- shodh-memory v0.1.6
- Source files in `src/config.rs`

### Documented defaults (multiple locations):

**`src/config.rs` lines 190-193 (comments):**
```rust
/// Rate limit: requests per second (default: 1000 - LLM-friendly)
pub rate_limit_per_second: u64,

/// Rate limit: burst size (default: 2000 - allows rapid agent bursts)
pub rate_limit_burst: u32,
```

**`src/config.rs` lines 407-408 (`print_env_help()`):**
```
SHODH_RATE_LIMIT       - Requests per second (default: 1000)
SHODH_RATE_BURST       - Burst size (default: 2000)
```

**`.env.example` lines 31-32:**
```
# SHODH_RATE_LIMIT=1000
# SHODH_RATE_BURST=2000
```

### Actual defaults (`src/config.rs` lines 235-236):

```rust
rate_limit_per_second: 4000,
rate_limit_burst: 8000,
```

## Expected Behavior

Documentation matches implementation:

| Parameter | Documented | Actual | Should Be |
|-----------|-----------|--------|-----------|
| `rate_limit_per_second` | 1000 | 4000 | 4000 |
| `rate_limit_burst` | 2000 | 8000 | 8000 |

## Suggested Fix

Update these locations:
1. `src/config.rs` line 190: Change "default: 1000" to "default: 4000"
2. `src/config.rs` line 193: Change "default: 2000" to "default: 8000"
3. `src/config.rs` line 407: Change "(default: 1000)" to "(default: 4000)"
4. `src/config.rs` line 408: Change "(default: 2000)" to "(default: 8000)"
5. `.env.example` lines 31-32: Update commented values

PR #285, opened 2026-05-21
feat: port veld field-encryption, HTTP embedder, constant-time compare

Portll:
## Summary

Ports four features from the sibling **veld** codebase into shodh-memory,
keeping module source **identical where practical** so the two projects can
stay in sync ("siblings"). Every change is secure-by-default and backward
compatible — nothing activates unless explicitly configured.

### 1. Field-level encryption at rest — `src/encryption.rs` (new)
AES-256-GCM authenticated encryption of the memory `content` field before it
hits RocksDB.
- Keyed by `SHODH_ENCRYPTION_KEY` — 64-char hex or 44-char base64 (32 bytes).
- **Unset = disabled**, content stored as plaintext (backward compatible).
- Wired via a process-global encryptor in `storage.rs`: `encode_memory()`
  encrypts on every write path (6 call sites + migration re-encode);
  `deserialize_memory()` decrypts on every read path (single chokepoint).
- A 4-byte `ENC\0` marker makes reads self-describing — pre-existing plaintext
  memories keep working unchanged, and are upgraded to ciphertext on next write.
- Key material held in `Zeroizing` / wiped on drop.

### 2. HTTP embedding backend — `src/embeddings/http_embedder.rs` (new)
`HttpEmbedder` — an OpenAI-compatible `/v1/embeddings` client (LM Studio,
Ollama, vLLM). Implements the existing `Embedder` trait; registered/exported
from `embeddings/mod.rs` but not wired into default selection (no behavior
change). Configured via `SHODH_EMBEDDING_API_{URL,MODEL,KEY,TIMEOUT_MS}`.

### 3. `subtle` constant-time comparison — `src/auth.rs`
`constant_time_compare` now uses the audited `subtle::ConstantTimeEq` with
equal-length normalisation, replacing the hand-rolled accumulator. Same
signature; all existing auth tests pass unchanged.

### 4. `zeroize`
Encryption key material wiped on drop (used by `encryption.rs`).

## Dependencies

New: `zeroize`, `subtle`. (`aes-gcm`, `base64`, `hex`, `ureq` were already
present — same versions veld uses.)

## Sibling parity

`encryption.rs` and `http_embedder.rs` are byte-identical to veld's modules
except for the `VELD_*` → `SHODH_*` env-var names and doc headers. Their veld
unit tests are carried over verbatim. The storage *wiring* necessarily differs
(shodh uses a postcard/SHO-envelope serialization layer; veld uses bincode) —
a process-global encryptor keeps shodh's wiring to one decode chokepoint.

## Verification

- `cargo fmt` passes — every changed/new file is syntactically valid.
- Full `cargo build` / `cargo test` was **not** run locally: the
  `librocksdb-sys` C++ build is environment-blocked on this machine. **CI must
  run the authoritative build + test.** The two new modules carry comprehensive
  unit tests (crypto round-trip, marker detection, wrong-key, key decoding;
  embedder config + dimension default).
- API usage (`aes-gcm`, `ureq` 3.x, `subtle`) is identical to veld's
  compiling code and to existing shodh call sites (`downloader.rs` uses the
  same `ureq` 3.x API).

## Follow-ups (out of scope)

- Wire `HttpEmbedder` into the embedder auto-selection (e.g. a `SHODH_EMBEDDER`
  env switch) — currently available as a library type only.
- A live encrypt/decrypt round-trip integration test against a real RocksDB
  store, once CI confirms the build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,3 +104,9 @@ pitch-deck-stata.md
 shodh-market.csv
 benchmarks/locomo_test2.json
 seps/mcp-sep-mif.md
+
+# Prebuilt CI tools — downloaded fresh at workflow time, never committed.
+# See .github/workflows/mcp-registry.yml which downloads mcp-publisher itself.
+/mcp-publisher
+/mcp-publisher.exe
+/mcp-publisher*.tar.gz
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4048,6 +4048,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2",
+ "subtle",
  "tantivy",
  "tar",
  "tempfile",
@@ -4066,6 +4067,7 @@ dependencies = [
  "uuid",
  "yake-rust",
  "zenoh",
+ "zeroize",
  "zip",
 ]
 
@@ -6137,6 +6139,20 @@ name = "zeroize"
 version = "1.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b97154e67e32c85465826e8bcc1c59429aaaf107c1e4a9e53c8d8ccd5eff88d0"
+dependencies = [
+ "zeroize_derive",
+]
+
+[[package]]
+name = "zeroize_derive"
+version = "1.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "85a5b4158499876c763cb03bc4e49185d3cccbabb15b33c627f7884f43db852e"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.110",
+]
 
 [[package]]
 name = "zerotrie"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -194,6 +194,10 @@ hex = "0.4"
 
 # MIF encryption (AES-256-GCM)
 aes-gcm = "0.10"
+# Zeroize key material on drop (field-level encryption)
+zeroize = { version = "1", features = ["derive"] }
+# Constant-time equality for API key comparison
+subtle = "2.6"
 
 # External integrations (SHO-40)
 hmac = "0.12"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -298,6 +298,10 @@ python = ["pyo3", "numpy"]
 # Optional: LLM-based query parser (future: requires API key)
 llm-parser = []
 
+# Optional: HTTP-backed OpenAI-compatible embedder.
+# Disabled by default because it can send memory text to an external service.
+http-embedder = []
+
 # Optional: Zenoh transport for robotics/ROS2 (pure Rust, cross-platform)
 zenoh = ["dep:zenoh"]
 
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,18 @@
+shodh-memory
+Copyright (c) shodh-memory contributors
+
+This product is licensed under the Apache License, Version 2.0 (see LICENSE).
+
+----------------------------------------------------------------------
+Third-party / cross-project attributions
+----------------------------------------------------------------------
+
+The following source files were contributed by Portll, the original author,
+under the Apache License 2.0. Portll also maintains parallel implementations of
+this code in its `veld` project (licensed BUSL-1.1) at commit
+dbc9036d5c29614eba9a6d43032038632d5612f9. As the copyright holder, Portll
+contributes these files to shodh-memory under Apache-2.0; no BUSL-1.1 terms
+attach to the copies in this repository.
+
+  - src/encryption.rs              (AES-256-GCM record-level encryption)
+  - src/embeddings/http_embedder.rs (OpenAI-compatible HTTP embedding client)
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -29,3 +29,85 @@ If you discover a security vulnerability in shodh-memory, please report it priva
 - Disclose publicly before a fix is available
 
 We appreciate responsible disclosure.
+
+---
+
+# Encryption-at-rest & embedder egress — threat model
+
+This section states precisely what the optional at-rest encryption and the HTTP
+embedder do and do **not** protect, so the guarantees are not over-read from a
+feature name.
+
+## Encryption at rest (`SHODH_ENCRYPTION_KEY`)
+
+When `SHODH_ENCRYPTION_KEY` is set (64-char hex or 44-char base64 → 32 bytes),
+each `Memory` record is encrypted with AES-256-GCM **after** serialization and
+stored as `[ENC\0 marker][12-byte nonce][ciphertext+tag]` in the primary column
+family. When unset, records are stored as plaintext (backward compatible);
+pre-existing plaintext records remain readable and upgrade to ciphertext on next
+write.
+
+### Covered
+
+- The full serialized `Memory` record in the primary CF: content, summary, tags,
+  entity refs, embeddings, and all other serialized fields are opaque on disk.
+- Ciphertext authentication (GCM tag): tampering/corruption surfaces as a
+  decryption **error**, not silent garbage.
+
+### NOT covered (known, deliberate gaps)
+
+- **Secondary index (`memory_index` CF).** Derived lookup keys are stored in
+  **plaintext**: `tag:<tag>:<id>`, `entity:<name>:<id>`, and
+  date/type/importance/geo/action keys. An on-disk index scan can confirm "a
+  memory tagged X (or referencing entity Y, or from date Z) exists" **without
+  touching ciphertext**. The *values* live in the encrypted record; the
+  *existence and key terms* do not. Blinding the index is tracked as separate,
+  deferred work.
+- **Sibling column families** (feedback, files, prospective, todos) and any
+  separately-keyed embedding blobs are not routed through the encryptor and are
+  stored plaintext.
+- **Memory-resident plaintext.** Decrypted content lives in process memory while
+  in use; this is at-rest protection only.
+
+### Key-loss / key-mismatch protection
+
+On first encrypted open, a non-secret 4-byte fingerprint (`SHA-256(key)[..4]`)
+is written to a sentinel entry in `memory_index`. On every subsequent open:
+
+- stored fingerprint present + configured key **mismatches** → hard error,
+  refuse to serve (prevents serving ciphertext-as-plaintext after a key swap);
+- stored fingerprint present + **no** key configured → hard error;
+- no stored fingerprint + key configured → fingerprint is recorded.
+
+A malformed `SHODH_ENCRYPTION_KEY` is a **hard failure** (panic at first use),
+not a silent fallback to plaintext.
+
+> The fingerprint is a tripwire, not an authenticator: non-secret, only 4 bytes.
+> It defends against accidental key loss/mismatch, **not** against an adversary
+> who can already write to the database (game-over regardless).
+
+### Key rotation — UNSUPPORTED in-process
+
+The process-global encryptor is initialised once (`OnceLock`) and cannot change
+without a restart. There is **no online key-rotation path**. To rotate: stop the
+service, run an offline re-encrypt (decrypt-with-old → encrypt-with-new over
+every record, reset the sentinel), then restart with the new key. An offline
+`rotate-key` subcommand is not yet implemented.
+
+## HTTP embedder egress (`http-embedder` feature, off by default)
+
+The HTTP embedder compiles only under the `http-embedder` Cargo feature. When
+enabled, `SHODH_EMBEDDING_API_URL` is validated before use by a self-contained
+guard in the embedder module (it does not depend on the integration-URL helpers
+owned by PR #284):
+
+- plain-HTTP to a **non-localhost** host is rejected by default (falls back to
+  the localhost default), because the embedder POSTs user text to that URL;
+- opt back in with `SHODH_ALLOW_INSECURE_REMOTE_EMBEDDER=true`.
+
+This blocks the obvious env-injection exfiltration / SSRF vector
+(`http://169.254.169.254/...`). It does **not** cover `https://` URLs to
+internal/link-local hosts, or DNS rebinding (a hostname resolving to an internal
+address), because no name resolution or address-range check is performed. Treat
+the embedder URL as trusted configuration. Availability checks use `HEAD` (no
+billable embedding request) and are cached for 30s.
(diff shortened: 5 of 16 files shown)


Portll:
## Local verification complete

The "CI must verify" caveat in the PR body is now resolved. With the
build environment restored:

**Build + tests** (target = `C:/sm-t` short-path, default parallelism)
- `cargo check --lib --bins` — exit 0, 4m02s, no errors / no warnings
- `cargo test --lib encryption::tests` — **12/12 pass** (round-trip, marker
  detection, wrong-key, truncated, unicode, hex/base64 key decode, `from_env`)
- `cargo test --lib http_embedder::tests` — **2/2 pass** (config defaults,
  dimension default)
- `cargo test --lib auth::tests::constant_time` — **6/6 pass** (the
  `subtle` swap preserves every existing case incl. the
  length-multiple-of-256 regression)
- `cargo clippy --lib --bins --no-deps` — **0 warnings** (4 pre-existing
  warnings in unrelated files were cleared in a237efa)

**Security review** of the full diff: no HIGH or MEDIUM-confidence
findings. Crypto primitives used correctly (per-call `OsRng` nonce,
AEAD tag verification surfaces `Err`, `Zeroizing` key + explicit `Drop`),
no user-controlled URL/header surface in the HTTP embedder, no oracle
on decrypt failure.

**Branch state** (in addition to the original port):
- `a237efa` clear pre-existing clippy warnings
- `8d9ede4` untrack accidentally-committed `mcp-publisher` binaries (25.5 MB)
- `b65b37a` companion `.gitignore` patterns

Deliberate follow-ups still flagged in the PR body: an end-to-end
`SHODH_ENCRYPTION_KEY` round-trip integration test against a real
RocksDB store, and wiring `HttpEmbedder` into the embedder
auto-selection.

varun29ankuS (changes requested):
Thanks John — appreciate the work, and the binary cleanup (`mcp-publisher.{exe,tar.gz}` deletion + `.gitignore`) is a nice side benefit, shrinks the working tree by ~25 MB. The AES-GCM nonce strategy (96-bit `OsRng` per encryption) is right, the `Zeroizing<[u8; 32]>` key handling is right, the `ENC\x00` marker is a clever discriminator, the `subtle::ConstantTimeEq` swap is the right call in principle, and the seven `encode_sho` call sites in `storage.rs` + the one in `migration.rs` are all converted (I checked).

That said — encryption-at-rest is high blast radius and there are several defects I'd want fixed before this merges. Some are correctness; some are about the gap between what the PR description claims and what the diff actually delivers — easy for a future reader (or our own future selves) to misread otherwise.

### Blockers

**1. Encryption scope is narrower than the PR description implies.**

`encode_memory` in `src/memory/storage.rs` only encrypts `memory.experience.content`:

```rust
let encrypted_bytes = enc.encrypt_content(&memory.experience.content)?;
let encoded = base64::engine::general_purpose::STANDARD.encode(&encrypted_bytes);
let mut encrypted_memory = memory.clone();
encrypted_memory.experience.content = encoded;
crate::serialization::encode_sho(&encrypted_memory)
```

Everything else in the serialised `Memory` — `metadata`, `tags`, `entity_refs`, `summary`, the embedding vector — stays plaintext on disk. The embedding alone is semantically invertible (Song et al. 2020, "Information Leakage in Embedding Models"), `summary` is often a verbatim subset of content, and `entity_refs` are extracted from content *before* encryption — so an index query like `entity_refs = ["foo"]` returns hits without touching the ciphertext path at all.

Calling this "field-level encryption" is technically accurate but oversells the threat model. Two options:

- **Expand**: encrypt the entire serialised `Memory` blob (encrypt after `encode_sho`, with the marker living outside the SHO envelope). Single chokepoint, everything opaque on disk.
- **Document honestly**: keep the current scope but rename/relabel as "content-field defence-in-depth" and add a `SECURITY.md` note enumerating what *isn't* covered.

**2. Base64-stuffing the ciphertext into a `String` field is expensive.**

The pipeline in `encode_memory` is: plaintext → AES-GCM bytes → **base64** → `String` field → postcard. For a 1 KB memory this inflates on-disk size by ~33% and kills RocksDB block compression (base64's high-entropy alphabet doesn't compress well). The reason in the code comment is that ciphertext isn't valid UTF-8 — fair, but the right fix is `serde_bytes::ByteBuf` (or a `Vec<u8>` slot on `Experience`), or encrypt-after-serialise which sidesteps the string-vs-bytes question entirely.

**3. Key-loss is silent data-loss with no recovery surface.**

Scenario: operator sets `SHODH_ENCRYPTION_KEY`, writes 10k records, restarts without the env var. `FieldEncryptor::from_env()` returns `Ok(None)`, the `OnceLock` initialises to `None`, encryption is "disabled" — and reads of old records flow through `decrypt_memory_content`, hit the `match encryptor() { None => return, ... }` arm, and return base64-of-ciphertext as the content field. No error, no log, no metric. The caller thinks it has a valid memory.

This is the worst possible at-rest failure mode. Minimum fix: stamp a key fingerprint (first 4 bytes of SHA-256(key)) into a sentinel CF entry on first encrypted write; on startup, if the CF entry exists but the env-derived key doesn't match (or is absent), log CRITICAL and refuse to serve those records (or refuse to start).

Also: there's no rotation path. `OnceLock` forecloses changing the key in-process. Document that explicitly or add a `rotate` CLI subcommand.

**4. Malformed key is swallowed into "disabled" state.**

In `encryptor()`:

```rust
.get_or_init(|| match crate::encryption::FieldEncryptor::from_env() {
    Ok(enc) => enc,
    Err(e) => {
        tracing::error!("SHODH_ENCRYPTION_KEY is set but invalid — content encryption DISABLED: {e}");
        None
    }
})
```

Operator typoes the key → server starts cleanly → writes plaintext → next time they fix the typo, the database has mixed plaintext-and-encrypted records with no mechanism to tell them apart (the `ENC\x00` marker only differentiates *if* you assume the operator's plaintext never starts with `ENC\x00`-encoded base64). This should be a hard `panic!` / `process::exit(1)` in production mode. Match the philosophy `auth.rs` already uses for missing auth config.

### Other defects (please address before merge)

- `decrypt_memory_content` in `src/memory/storage.rs` logs on decryption failure but returns `()` with `memory.experience.content` still set to base64-ciphertext. Caller (`deserialize_memory`) has no signal that decryption failed and serves garbage downstream. Should propagate the error.
- Test `from_env_returns_none_when_unset` in `src/encryption.rs` calls `std::env::remove_var("SHODH_ENCRYPTION_KEY")` with no `ENV_LOCK`. `cargo test` runs in parallel and `auth.rs` already solved this with the `ENV_LOCK` pattern — please follow that convention or `#[ignore]` the test. (Rust 1.85+ marks `env::remove_var` `unsafe` in multithreaded contexts.)
- `HttpEmbedder::is_available()` in `src/embeddings/http_embedder.rs` sends an actual `EmbeddingRequest { input: "test", ... }` to the configured endpoint. On any paid API (OpenAI, Cohere) that's a billable request every call, with no rate limit or caching. Use a `HEAD /` / dedicated health endpoint, or cache the result for N seconds.
- **SSRF surface in `HttpEmbedder`**: `SHODH_EMBEDDING_API_URL` is read raw via `std::env::var` with no scheme/host validation. With a compromised env (`SHODH_EMBEDDING_API_URL=http://169.254.169.254/latest/meta-data/...`), the embedder will POST batches of user text to whatever URL was injected — both an SSRF and a data-exfil channel. PR #284 introduces `is_insecure_remote_url` / `resolve_api_url_override` / `SHODH_ENFORCE_HTTPS` for exactly this; please use it. This is one reason to land #284 first.
- `constant_time_compare` in `src/auth.rs`: the length-mismatch branch runs `b_bytes.ct_eq(b_bytes)` as a dummy, whose cost is `len(b)`. `b` is the user-supplied key, so its length is attacker-controlled and the dummy's cost varies with it. The hand-rolled implementation you're replacing always iterated `max(len_a, len_b)`. Either pad-to-max before `ct_eq`, or enforce a single configured key length upfront and reject anything else (cleaner — API keys are length-known-public anyway).
- `HttpEmbedder` is `pub use`-exported from `src/embeddings/mod.rs` but never wired into the selection logic. Shipping unused crypto-adjacent code that calls external HTTP servers is an audit liability. Either wire it behind a `SHODH_USE_HTTP_EMBEDDER=true` toggle (with the SSRF guards from #284) or put it behind a `#[cfg(feature = "http-embedder")]` flag.

### Veld attribution

The PR description says the modules are byte-identical to veld's. Apache 2.0 requires retaining `NOTICE` attribution; I didn't see an upstream-attribution header in `encryption.rs` or `http_embedder.rs`. Could you add a `// vendored from veld @ <commit-sha>` header (or a `NOTICE` entry) and confirm veld's license is compatible with shodh-memory's Apache 2.0?

### Order

Could we land #284 first? The SSRF guard and error-sanitisation helpers it adds are direct dependencies for #285's `HttpEmbedder` and the crypto error paths.

### Things that are right

For the record so you know what I checked and was happy with:

- AES-256-GCM construction in `encryption.rs`: 96-bit `Aes256Gcm::generate_nonce(&mut OsRng)` per encryption. Collision-safe up to ~2^48 encryptions (catastrophic at ~2^32) — well within real workload.
- `Zeroizing<[u8; 32]>` key holding + manual `Drop` impl wiping. `aes-gcm 0.10`'s `Aes256Gcm` also `ZeroizeOnDrop`s its key schedule.
- `ENC\x00` 4-byte marker: works because serde-encoded UTF-8 strings can't contain literal NUL the way base64-of-binary can. (Worth a doc-comment explaining *why* it works — non-obvious from the constant alone.)
- All seven `encode_sho` call sites in `storage.rs` and the one in `migration.rs` are converted to `encode_memory`. Coverage verified.
- `Embedder` trait conformance on `HttpEmbedder` is clean; `Send + Sync` bounds satisfied via `ureq::Agent`.

Happy to chat through any of this if it's easier than back-and-forth on the PR.

Portll:
Thanks for the thorough review, @varun29ankuS — every point was actionable. Pushed `916127c` on top of the earlier hardening commit. Note: the build is locally verifiable after all — the "rocksdb environment-blocked" was just a missing `LIBCLANG_PATH` (`LIBCLANG_PATH=$(brew --prefix llvm)/lib cargo check` builds clean), and the 26 auth tests pass.

### Fixed in `916127c`

- **`constant_time_compare` length leak (your "other defects" #5).** The dummy `b_bytes.ct_eq(b_bytes)` — whose cost was the attacker-controlled `len(b)` — is gone. It now enforces equal length (API-key length is public) then `subtle::ct_eq`, which also closes the DoS where a multi-MB "key" forced a multi-MB compare. The misleading "does not leak length" comment is deleted. All 6 constant-time tests pass, including the 256-multiple regression.
- **`is_available` billable request.** Now a `HEAD` + 30s availability cache; no more `input:"test"` POST per call.
- **SSRF + the #284 ordering.** Instead of blocking on landing #284 first, the embedder now carries a *self-contained* insecure-remote-URL guard (opt out via `SHODH_ALLOW_INSECURE_REMOTE_EMBEDDER`), and I **reverted** the `is_insecure_remote_url` / `resolve_api_url_override` copies this PR had added to `integrations/mod.rs`. That leaves #284 the sole owner of those symbols, so the two PRs no longer collide on a duplicate definition / duplicate `mod tests`. (Documented caveat: this blocks plain remote-HTTP egress; it does not cover `https://`-to-internal or DNS-rebind.)
- **Encryption scope honesty (your blocker #1).** You're right that "opaque at rest" over-sold it. The docstring is narrowed and a new `SECURITY.md` threat-model section enumerates what is *not* covered — most importantly that the `memory_index` CF stores plaintext `tag:` / `entity:` / `date:` … keys, so an index scan reveals existence/terms without touching ciphertext. The same section documents the key-rotation limitation and the embedder egress posture.
- **Attribution.** Apache-2.0 contribution headers on `encryption.rs` and `http_embedder.rs` + a `NOTICE`. On the license question: shodh-memory is the Apache-2.0 base; Portll authored these and contributes them back under Apache-2.0 — the veld cross-reference is informational and no BUSL-1.1 terms attach.

(Your blockers #2 base64-inflation, #3 silent key-loss sentinel, #4 malformed-key panic, and defects #6 ENV_LOCK / #10 feature-gating were already in the prior commit.)

### Deferred to a separate PR

You correctly identified that real confidentiality needs more than content encryption. Rather than bolt partial index-blinding onto this PR — the order-dependent indexes (date range, importance threshold, geo proximity, episode sequence) can't be naively HMAC-blinded without breaking range queries, and it needs a destructive index rebuild — we're doing a proper **encryption-at-rest v2** as its own PR: envelope KEK/DEK with a keystore, passphrase + OS-keychain + KMS unseal, key rotation, a recovery code, and full index blinding. Glad to loop you in on that design.

The same hardening was mirrored into the **veld** sibling (D5/D7/D8/B3); `constant_time_compare` was deliberately left untouched there since veld already has the correct max-length implementation — porting this PR's version would have reintroduced the bug.

Re the base conflict flagged on the PR: I'll rebase once #284 lands.

varun29ankuS:
Verified `916127c` — the constant-time fix is right: early length-reject (key length isn't secret) + `subtle::ct_eq` on equal-length content, dummy compare gone, DoS closed, docstring honest. 👍 Same for the `HEAD` availability cache and the SSRF de-dup that gives #284 sole ownership of `is_insecure_remote_url` / `resolve_api_url_override`.

Sequencing from my side:
1. I'll merge **#284** (green across all platforms) so you can rebase onto it.
2. After the rebase, this branch needs a CI run — it currently shows no checks and `mergeState=DIRTY`, so the local "26 auth tests pass" isn't confirmed on CI yet. Once it's green I'll do the final review pass.

On **encryption-at-rest v2**: yes, loop me in. The index-blinding constraint you flagged (order-dependent date/importance/geo/episode indexes can't be naively HMAC-blinded without breaking range queries) is the crux — I'd want to see how the envelope KEK/DEK + keystore handles rotation against an *existing* store before committing to the destructive rebuild. Send a design doc or draft PR.

Thanks for the thorough turnaround.

Portll:
Superseded by #318 (encryption-v2).

PR #318 builds on this branch's work and replaces the v1 single-key
field-encryption approach with the keystore-based encryption-v2 core
(XChaCha20-Poly1305 records, Argon2id-sealed keystore with passphrase/KMS
unseal + rotation, HMAC index blinding, oblivious-access seam). The standalone
v1 FieldEncryptor introduced here has since been removed entirely in #318 —
record storage, slow-store payloads, and TOTP secrets all run on the keystore.

Closing in favour of #318.

PR #288, opened 2026-05-25
chore: clean sweep of clippy 1.94 warnings across lib + tests

varun29ankuS:
## Summary

Clean sweep of clippy 1.94 errors and warnings across both production lib and test code so all three gates pass with `-D warnings`:

- `cargo check --lib`
- `cargo clippy --lib -- -D warnings`
- `cargo clippy --tests --lib -- -D warnings`

Originally cut as a small pre-blocker for MCAP Phase 1 commit-1 DoD; expanded by request to cover the full test surface.

### Production lib — behavior-neutral refactors

- **`src/memory/hybrid_search.rs`** — Removed dead `get_content: F` closure parameter threaded through 5 wrapper layers (`search`, `search_with_ic_weights`, `search_with_ic_weights_and_phrases`, `search_with_dynamic_weights`, `search_with_dynamic_weights_pool`). The bottom layer ignored it (`_get_content`); the parameter served no purpose downstream. Fixes `clippy::too_many_arguments` (8/7).
- **`src/memory/mod.rs`** — Removed the `get_content,` arg from the lone external caller. Kept the closure declaration itself because it is still consumed by the prospective-signals boost (Layer 4.7) at line 2590 — added a comment narrowing its documented purpose.
- **`src/memory/query_parser.rs`** — `.iter().any(|x| *x == s.as_str())` → `.contains(&s.as_str())` (`POLAR_QUESTION_LEADERS` is `&[&str]`). Fixed doc-list 8-space continuation → 4-space.
- **`src/embeddings/chunking.rs`** — `x >= 5 && x <= 15` → `(5..=15).contains(&x)`.
- **`src/recall_harness/metrics.rs`** — Redundant closure `|n| Uuid::from_u128(n)` → `Uuid::from_u128`.
- **`src/token_estimation.rs`** — `(x * 3 + 1) / 2` → `(x * 3).div_ceil(2)`.

### Test code — cfg(test) modules inside src/

- **`src/auth.rs`** (8 `#[tokio::test]` fns) and **`src/middleware.rs`** (2 fns) — Added `#[allow(clippy::await_holding_lock)]` narrowly per test function. The `ENV_LOCK: Mutex<()>` is held across `.await` by design — it serializes tests that mutate process-global env vars (`SHODH_API_KEYS`, `SHODH_ENV`). Releasing earlier or swapping to `tokio::sync::Mutex` would reintroduce the cross-test env race the lock exists to prevent.
- **`src/handlers/state.rs`, `src/memory/graph_retrieval.rs`, `src/memory/types.rs`** — Converted `assert!(CONSTANT > 0.6, ...)` style runtime assertions on `const` values to `const _: () = assert!(...)`. These now fail at **compile time** instead of test time — strictly stronger. Tradeoff: runtime `{}` format args dropped (compile-time errors already point to the exact line and constant, no `format!` allowed in const-eval).
- **`src/handlers/test_helpers.rs`** — Added `impl Default for TestHarness { fn default() -> Self { Self::new() } }`.
- **`src/memory/replay.rs`** — `vec![(...)]` → `[(...)]` for one-shot literal.
- **`src/memory/graph_retrieval.rs`** — Same `useless_vec` cleanup at 5 sites.

### Test code — tests/ (15 files)

- **Dead-code allows** (`#[allow(dead_code)]`) added narrowly per-function to shared NER/setup helpers that are copy-pasted across test files (`setup_fallback_ner`, `create_experience_with_ner`, `create_entity_from_ner`, `ner_type_to_label`, `create_rich_experience_with_ner`, `create_conversation_experience`, `benchmark_operation`, `authed_put`, `authed_patch`, `INDEX_OP_MAX_MS`). Each test binary instantiates only the helpers it actually invokes, but the modules share a common copy-pasted block; deletion would force a per-file diff against the others. `#[allow]` is the honest fix.
- **`&[x.clone()]` → `std::slice::from_ref(&x)`** across `tests/brutal_stress_tests.rs`, `tests/memory_persistence_tests.rs`, `tests/hebbian_learning_tests.rs`, `tests/adaptive_memory_tests.rs` — eliminates redundant allocation in tests that pass a single-element slice.
- **Tautological `assert!(x >= 0)` on `usize`/`u32`/`u64` deleted** in `tests/brutal_stress_tests.rs`, `tests/memory_tiering_tests.rs`, `tests/adaptive_memory_tests.rs`, `tests/query_filter_tests.rs`. Where the surrounding call was a side-effect probe (e.g., `graph_stats()`), the call was preserved with an explanatory comment so the lint signal isn't silenced at the cost of the actual test intent.
- **`importance >= 0.0 && importance <= 1.0` → `(0.0..=1.0).contains(&importance)`** for legitimate float bounds checks (these are NOT tautological).
- **`format!` inside `.expect()` → `.unwrap_or_else(|_| panic!(...))`** in `tests/brutal_stress_tests.rs` and `tests/memory_persistence_tests.rs` — avoids eager allocation of the panic string on the hot success path.
- **`assert!(... > 0.0)` on imported constants → `const _: () = assert!(...)`** in `tests/query_filter_tests.rs`.

## Adversarial notes

- **Lesson from mid-fix bug:** my first pass at the `get_content` dead-parameter removal also deleted the closure construction in `memory/mod.rs`, breaking compilation because the closure is still consumed downstream by the prospective-signals layer. The parameter was dead in the search chain; the closure itself was not. Re-verified with `cargo check` before claiming the fix was complete. Pattern to watch: dead parameters at one level can mask still-live consumers elsewhere in the outer scope.
- **`await_holding_lock` allows are narrowly scoped per `#[tokio::test]` fn**, never module-wide. If you grep them out later, audit each — they exist because env-var serialization needs the lock across the request future. Removing the lock without a different serialization mechanism will reintroduce flaky test failures under `cargo test --jobs > 1`.
- **`const _: () = assert!(...)` over runtime `assert!`** chosen because the asserted value IS a `const`. Compile-time failure is strictly stronger than test-time failure and runs even with `#[ignore]`. Cost: lost runtime `{}` interpolation, but the failing constant's name and line are visible in the compiler error anyway.
- **Tautological asserts on unsigned types were deleted, not silenced.** `assert!(usize_val >= 0)` carries zero information. Where a message was potentially load-bearing for debugging the surrounding behavior, the side-effect call was preserved with a comment explaining why the original assertion was vacuous. No behavioral change.
- **Float `>= 0.0` bounds checks were preserved** (and modernized to range-contains where idiomatic). Only unsigned `>= 0` checks were deleted.
- **No `#[allow(...)]` was added at module/crate scope.** All allows are narrowly per-function or per-item.
- **Not in scope:** functional changes, dependency bumps, test additions, behavior changes. This PR is pure hygiene against clippy 1.94's expanded lint set.

## Test plan

- [ ] `cargo check --lib` passes
- [ ] `cargo clippy --lib -- -D warnings` passes
- [ ] `cargo clippy --tests --lib -- -D warnings` passes
- [ ] CI green
- [ ] (manual spot-check) at least one test from `tests/brutal_stress_tests.rs` runs successfully under `cargo test --test brutal_stress_tests` — confirms tautological assertion deletions didn't accidentally remove a load-bearing side effect

## Follow-up

This unblocks `feat/mcap-ingest` (MCAP Phase 1) whose commit-1 DoD requires `cargo clippy --features mcap --lib -- -D warnings` to be clean.

### DIFF
--- a/src/auth.rs
+++ b/src/auth.rs
@@ -470,6 +470,7 @@ mod tests {
     // ── AuthError JSON response shape ──
 
     #[tokio::test]
+    #[allow(clippy::await_holding_lock)]
     async fn auth_error_response_is_valid_json() {
         let _guard = ENV_LOCK.lock().unwrap();
         clear_auth_env();
@@ -485,6 +486,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[allow(clippy::await_holding_lock)]
     async fn missing_key_dev_message_includes_help() {
         let _guard = ENV_LOCK.lock().unwrap();
         clear_auth_env();
@@ -505,6 +507,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[allow(clippy::await_holding_lock)]
     async fn invalid_key_dev_message_includes_help() {
         let _guard = ENV_LOCK.lock().unwrap();
         clear_auth_env();
@@ -520,6 +523,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[allow(clippy::await_holding_lock)]
     async fn missing_key_prod_message_is_terse() {
         let _guard = ENV_LOCK.lock().unwrap();
         clear_auth_env();
@@ -536,6 +540,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[allow(clippy::await_holding_lock)]
     async fn invalid_key_prod_message_is_terse() {
         let _guard = ENV_LOCK.lock().unwrap();
         clear_auth_env();
@@ -561,6 +566,7 @@ mod tests {
     // ── Query parameter auth (WebSocket fallback) ──
 
     #[tokio::test]
+    #[allow(clippy::await_holding_lock)]
     async fn auth_middleware_accepts_query_param_for_websocket() {
         use axum::body::Body;
         use axum::http::Request as HttpRequest;
@@ -594,6 +600,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[allow(clippy::await_holding_lock)]
     async fn auth_middleware_ignores_query_param_without_websocket_upgrade() {
         use axum::body::Body;
         use axum::http::Request as HttpRequest;
@@ -626,6 +633,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[allow(clippy::await_holding_lock)]
     async fn auth_middleware_rejects_invalid_websocket_query_param() {
         use axum::body::Body;
         use axum::http::Request as HttpRequest;
--- a/src/embeddings/chunking.rs
+++ b/src/embeddings/chunking.rs
@@ -660,7 +660,7 @@ mod tests {
         // Code-like content with more punctuation
         let code = "fn main() { println!(\"hello\"); }";
         let tokens = estimate_tokens(code);
-        assert!(tokens >= 5 && tokens <= 15, "Code tokens: {}", tokens);
+        assert!((5..=15).contains(&tokens), "Code tokens: {}", tokens);
 
         // No whitespace (falls back to char-based)
         assert_eq!(estimate_tokens("abcdefgh"), 2); // 8 chars / 4 = 2
--- a/src/handlers/state.rs
+++ b/src/handlers/state.rs
@@ -3546,11 +3546,11 @@ mod tests {
 
     #[test]
     fn test_ner_confidence_floor_constant() {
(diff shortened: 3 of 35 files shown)


varun29ankuS:
Stale: this clippy-1.94 cleanup predates the recall campaign (+284/-273 from late May) and no longer applies cleanly; clippy passes on current main. Closing — a fresh sweep against current main can be opened if warranted.

PR #311, opened 2026-05-30
ci(locomo): manual LoCoMo-MC10 benchmark (held-out generalization check)

varun29ankuS:
@
## Why

You asked the right question: are the recall fixes *general* or fit to the 108-case smoke suite? recall.yml is what we **tune** against. This adds the held-out set we **validate** against.

**LoCoMo-MC10** is an external HuggingFace dataset (`Percena/locomo-mc10`, ~1,986 multi-turn dialogue multiple-choice questions) that none of the pipeline changes were diagnosed on. If a fix (e.g. #309 competition demote-not-delete) helps here too, that is real generalization, not overfitting.

## What

`workflow_dispatch`-only job: builds + serves `shodh-memory-server`, the eval ingests the LoCoMo dialogues via the HTTP API and answers each MC question from retrieved context (LLM judge), then uploads `locomo_results.json` + a step-summary accuracy table.

Inputs: `ref` (benchmark any commit — e.g. run pre-#309 vs main to measure the delta), `limit`, `provider`/`model`/`api_base`.

## ⚠️ Needs one secret you must add

The MC eval has no gold-memory-ID scoring — it needs an LLM to pick the answer. Add a key (free Groq tier works):
```
gh secret set LOCOMO_LLM_API_KEY --repo varun29ankuS/shodh-memory   # paste a Groq/OpenAI/Anthropic key
```
Then I can dispatch it (default: Groq `llama-3.1-8b-instant`, `--limit 200`). To prove #309 generalized, I will run it on `main` and on the pre-#309 commit and diff the accuracy.

No effect on existing CI — manual trigger only.
@

### DIFF
--- a/.github/workflows/locomo.yml
+++ b/.github/workflows/locomo.yml
@@ -1,0 +1,143 @@
+# LoCoMo-MC10 benchmark — INDEPENDENT generalization check.
+#
+# The L1 smoke suite (recall.yml) is what we tune against; this is the held-out
+# set we validate against. LoCoMo-MC10 is an external HuggingFace dataset
+# (Percena/locomo-mc10, ~1,986 multi-turn dialogue multiple-choice questions)
+# that the pipeline changes were NOT diagnosed on — so it answers "did a recall
+# change generalize, or just fit the 108 smoke cases?".
+#
+# Manual only (builds + serves shodh, ingests dialogues, calls an LLM judge —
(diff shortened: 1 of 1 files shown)


varun29ankuS:
Superseded: LoCoMo held-out benchmarking on main is covered by the locomo-recall workflow, alongside the new LongMemEval-S SOTA workflow. The MC10-specific manual job is redundant. Closing.

PR #314, opened 2026-05-30
perf(retrieval): flatten the double-RRF (vector + BM25 as independent legs)

varun29ankuS:
@
## What & why

Per-layer diagnostics (#308) proved the nested `RRF(graph, RRF(bm25, vector))` **loses recall**: a hit ranked #1 by vector vanished the moment BM25 fusion was added (smoke-033 `ssm-010` "parking_lot": `vamana_only` recall 1.0 → `+bm25` **0.0**). RRF-averaging a strong single-source rank against a mediocre one buries it.

**Fix:** one flat RRF over three **independent** legs (graph + vector + bm25). The density weights already split into `(graph_w, semantic_w, linguistic_w)` summing to 1 → `graph→graph_w`, `vector→semantic_w`, `bm25→linguistic_w`. High YAKE discriminativeness still leans the vector/bm25 split toward BM25, preserving the prior dynamic-weight behaviour at the leg level. Below `PlusBm25` the vector leg carries the full non-graph weight (unchanged).

- New `hybrid_search::bm25_ranked` returns the IC/phrase-weighted BM25 leg alone.
- Layer 4 drops the nested `search_with_dynamic_weights_pool` call.

## Measurement (this is a ranking change — not asserting)
- Smoke gate (per-category + per-layer): expect smoke-033 to survive `+bm25` now.
- Held-out LoCoMo dispatch vs main, targeting **multi_hop** (weakest on both benchmarks: 0.29 LoCoMo / partial smoke).

Will not merge unless both move the right way.
@

### DIFF
--- a/src/memory/hybrid_search.rs
+++ b/src/memory/hybrid_search.rs
@@ -734,6 +734,33 @@ impl HybridSearchEngine {
     /// Variant of `search_with_dynamic_weights` that allows the caller to
     /// override the BM25 candidate pool depth.
     ///
+    /// IC-weighted BM25 ranking **without** the vector fusion — the BM25 leg on
+    /// its own, ranked by score descending and filtered to `min_bm25_score`.
+    ///
+    /// Lets the caller fuse BM25 as an INDEPENDENT leg in a flat RRF
+    /// (graph + vector + bm25) instead of nesting it inside a hybrid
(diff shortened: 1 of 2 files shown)


varun29ankuS:
Superseded: calibrated-max fusion (SHODH_FUSION_FLAT) shipped as the default fusion on main (#325), which subsumes the double-RRF flattening proposed here. Closing as obsolete.

PR #317, opened 2026-06-05
feat(eval): E7 fact-extraction quality harness

varun29ankuS:
## Summary
**E7 — the first harness that scores fact CORRECTNESS.** The other capability suites measure retrieval rank; the `+facts` layer contributes ~0, so a weak fact extractor previously read as a null, not a regression. This closes that hole.

## What it does
Plants memories with known gold facts (FactType-aligned, two mentions each) + no-fact distractors, forces `distill_facts(min_support=1, min_age_days=0)` to bypass the eval consolidation age gate, reads back `SemanticFact`s, and scores **precision / recall / F1 / dedup / per-type recall / confidence calibration**. Wired as `recall-eval --facts`.

`cargo check` green.

### DIFF
--- a/src/bin/recall_eval.rs
+++ b/src/bin/recall_eval.rs
@@ -241,6 +241,12 @@ struct Args {
     #[arg(long)]
     lineage: Option<PathBuf>,
 
+    /// E7 fact-extraction QUALITY: plant memories with known gold facts, force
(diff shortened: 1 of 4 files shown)


varun29ankuS:
Closing — stale (126 commits behind main). The fact-extraction quality harness can be recreated cleanly against current main if still wanted.

PR #318, opened 2026-06-05
feat(encryption-v2): crypto core — keystore, rotation, KMS, oblivious seam, PIR

Portll:
## Encryption-at-rest v2 — crypto core (port from veld)

Ports the storage-agnostic **encryption-v2** crypto core from Portll/veld
(`feat/encryption-v2`) into shodh-memory, adapted to `SHODH_*` env names. This is
the substantive, fully-tested foundation; the shodh-specific storage integration
is the documented next step (see below).

### Included (28 unit tests, all green)
- **`src/keystore.rs`** — envelope key hierarchy (a master key/KEK wraps per-epoch
  data keys/DEKs); Argon2id passphrase unseal; non-secret key fingerprint; the
  record-envelope primitive (`ENC\0|ver|epoch|nonce|ct`, epoch bound into GCM AAD);
  `IndexBlinder` (HMAC index-term blinding); passphrase + DEK **rotation** and a
  one-time **recovery code**; a `KmsClient` trait + `LocalAeadKms`
  (`SHODH_KMS_WRAP_KEY`) for **unattended unseal** (cloud KMS / OS keychain plug in
  by implementing the trait).
- **`src/oblivious.rs`** — `BlockStore` seam + `pad_to_bucket` mitigation
  (access-pattern threat model T2; Path-ORAM behind a future flag).
- **`src/pir.rs`** — 2-server information-theoretic PIR for exact-key private fetch
  (threat model T3); single-server cPIR (HE) documented as future, not fake-shipped.
- **`docs/encryption-v2-design.md`** — full design, T1/T2/T3 threat model, and the
  bifocal/overloop/breakers adversarial review.
- Adds the `argon2` dependency.

### Next step (storage integration)
Replacing the current single-key record encryption with the keystore, routing
serialize/decode through it, and applying index blinding at shodh's index sites —
mirroring the **proven, merged-on-branch** veld commits (record wiring + index
blinding + e2e RocksDB tests) on `Portll/veld:feat/encryption-v2`.

### Notes
- **Stacked on #285:** this branch builds on the #285 remediation (it needs the
  `zeroize`/`subtle`/`hmac` deps that PR adds). Review after #285 merges, or I'll
  rebase. The crypto modules themselves are independent and self-contained.


Portll:
## shipcheck + breakers — encryption-v2 crypto core

QC/QA + adversarial security pass on this branch.

### Fixed (pushed, 356847b)
- **Key-material hygiene (breakers):** `unwrap_key` / `unseal_with_kms` copied the decrypted KEK/DEK into the `Zeroizing` buffer but left the source `Vec` un-wiped, so key bytes lingered in freed heap (threat model T2: memory disclosure / core dump). Now zeroized after copy.

### Verified clean
- 28 module unit tests pass (22 keystore + 2 oblivious + 4 pir); fmt-clean; clippy-clean on keystore/oblivious/pir. (The 4 clippy lints from `cargo clippy` are pre-existing in `vector_db/distance_inline.rs`, unrelated to this PR.)
- AES-256-GCM with per-call random nonce, AEAD tag verified (decrypt surfaces `Err`); DEK wraps bind epoch into AAD (cross-epoch substitution blocked, tested); Argon2id 256 MiB / 3 passes; itPIR reconstruction + per-share uniform randomness; multi-wrap (passphrase / recovery / KMS) each unwraps the same fingerprinted KEK; no raw key material is serialized to the keystore.

### Residuals (documented by design — docs/encryption-v2-design.md §11)
- Record AAD binds epoch but not record-id → same-epoch ciphertext relocation needs DB write (already high-priv); id-binding is a follow-up.
- Index blinding leaks term equality/frequency; range fields are plaintext (B-3/B-4).
- Multi-wrap is weakest-link; keystore backup required (recovery code is the 2nd path); the 4-byte fingerprint is a tripwire, not an authenticator; auto epoch-rotation-by-count not yet wired.
- Storage integration is out of scope here (crypto core only).

Portll:
## Production-hardening update (crypto core)

Following a production-safety review, the encryption-v2 crypto core is hardened.
Landed here (and fully wired + e2e-tested on the veld sibling
`Portll/veld:feat/encryption-v2`):

- **H1 — atomic keystore persistence.** `Keystore::save_to_path` does temp +
  fsync + rename (+ `.bak`), so a crash never leaves a truncated/empty keystore
  (which would make the KEK — and all data — unrecoverable).
- **H2 — integrity + rollback.** Keystore carries a monotonic `generation` and an
  HMAC (keyed by a KEK-derived integrity key); `verify_integrity()` on load
  detects in-place tampering. (The DB rollback sentinel lands with H6 storage
  wiring.)
- **H3 — nonce ceiling removed.** Record encryption is now XChaCha20-Poly1305
  (192-bit random nonce); key-wrapping stays AES-256-GCM.
- Intermediate key-material buffers are zeroized after use.
- **H5 — `shodh-keyctl`** bin: `rotate-passphrase` / `rotate-dek` /
  `add-recovery-code`, persisted atomically.
- **H4 — confidentiality boundary** documented (what's encrypted vs the ANN
  vector index [plaintext by necessity], range index fields, sibling CFs) with a
  pre-production sign-off checklist (see veld `docs/ENCRYPTION-SECURITY.md`).

### Remaining before shodh is production-active (H6)
This PR is still the **crypto core + keyctl** — it is not yet wired into shodh's
storage, so shodh data isn't protected by v2 until the storage integration lands:
replace the single-key record encryptor with the keystore-based `RecordCryptor`,
add index blinding at shodh's index sites, and the DB rollback sentinel. veld's
`feat/encryption-v2` (record wiring + index blinding + e2e RocksDB tests) is the
exact, proven blueprint.

28 module unit tests pass (25 keystore + 2 oblivious + 4 pir... 31 total with pir/oblivious); shodh-keyctl builds; fmt + clippy clean on the v2 modules.

Portll:
## H6 landed — encryption-v2 is now active in shodh's storage

The crypto core is now **wired into the data path** (previously this PR was the
core + keyctl only):

- `encode_memory` / `deserialize_memory` use a process-global, keystore-derived
  **RecordCryptor** (XChaCha20-Poly1305 records) instead of the single-key
  `FieldEncryptor`. Unseal is `SHODH_MASTER_PASSPHRASE` (or `SHODH_KMS_WRAP_KEY`),
  not the old `SHODH_ENCRYPTION_KEY`.
- `MemoryStorage::new` loads/creates the keystore, **verifies its integrity MAC**,
  and enforces the **monotonic generation rollback sentinel** (`meta:keystore_generation`).
- **Index blinding live:** exact-match terms (tag / entity / external / parent /
  content_hash / episode / robot / mission / action) are HMAC-blinded at
  write/remove/query. Range/low-cardinality fields stay plaintext by design
  (see veld `docs/ENCRYPTION-SECURITY.md`).
- The old single-key e2e was replaced with a **v2 e2e** (round-trip, on-disk
  opacity, on-disk index blinding, wrong-passphrase hard error).

**Verified:** 17 storage unit tests (plaintext no-regression) + 3 e2e RocksDB
tests; lib builds clean; fmt clean.

### Status vs the production-safety review
All must-fix items are now addressed (H1 atomic persistence, H2 integrity+rollback,
H3 nonce ceiling, H4 boundary doc, H5 rotation CLI, H6 shodh activation). Remaining
are the documented, signed-off residuals (ANN vector index plaintext by necessity,
sibling CFs, index equality/frequency) — not silent gaps.

> Note: this PR remains stacked on #285 (it needs that PR's deps). Review after
> #285 merges, or I'll rebase.

Portll:
## Security audit (portll-security · breakers · shipcheck) — 2 findings fixed

A full adversarial pass over the v2 crypto + storage integration. Two real
findings, both fixed (`b93c03d`):

- **Argon2 param DoS (breakers/portll-security).** `KdfParams` (m/t/p cost) come
  from the *untrusted* `keystore.json`; a tampered `m_cost=u32::MAX` would force a
  multi-TB allocation / OOM at unseal, *before* any key check. `derive()` now
  rejects params above safe ceilings (4 GiB / 64 / 64).
- **Cross-keystore confusion (portll-security).** shodh's crypto is process-global;
  `V2_CRYPTO.set` silently ignored a second store with a *different* keystore →
  it'd reuse the first store's keys. Now the global carries the KEK fingerprint and
  a second, different keystore **fails loudly**. Added a test that asserts the
  guard fires.

### Verified sound (no change needed)
AEAD construction (XChaCha20-Poly1305 records, AES-GCM key-wrap, per-call random
nonce, tag-verified), epoch-bound AAD, Argon2id, integrity MAC over canonical
serialization, generation rollback sentinel, HMAC index blinding, key-buffer
zeroization, atomic keystore persistence.

### Documented residuals / scope (not bugs — see veld docs/ENCRYPTION-SECURITY.md)
- ANN vector index plaintext (inherent — similarity search); sibling CFs + range
  index fields plaintext (sign-off); index equality/frequency leak (HMAC).
- Record AAD binds epoch not record-id (relocation needs DB write); DB-side
  sentinels are plaintext (DB-write = out of scope).
- shodh's old `FieldEncryptor` (v1, #285) is now orphaned by v2 — remove when v1
  retires. veld's `FieldEncryptor` is still used by `slow_store` + TOTP secrets,
  which remain on the v1 single-key scheme (not v2-hardened).

**Tests:** 26 keystore + 17 storage unit + 4 e2e (round-trip, on-disk opacity,
index blinding, wrong-passphrase, cross-keystore guard); clippy + fmt clean.

varun29ankuS:
Thanks for this, @Portll — it's serious work, and it holds up to a serious review. We ran a multi-dimension security audit of the crypto core (envelope/AEAD, KDF/rotation, secret hygiene, PIR/oblivious, storage integration), and **the cryptography itself is sound**:

- Nonce management is correct — XChaCha20-Poly1305 for records, AES-GCM confined to key-wraps — so GCM nonce-reuse is structurally unreachable.
- Fail-closed envelope parsing; epoch-AAD-bound KEK→DEK wrapping; Argon2id at 256 MiB / t=3 (above OWASP); append-only rotation with **no** key-drop / data-loss path; secrets zeroized on the main paths; a textbook-correct CGKS information-theoretic PIR that's honestly scoped (no fake single-server cPIR, no deceptive ORAM stub). Supply chain is clean.

That's a strong foundation.

**Where we're headed with it:** we want to land this — as an **opt-in** capability. It genuinely sets shodh apart for security- and sovereignty-focused users (LLM-free **+** on-device **+** encrypted-at-rest is a combination our cloud-LLM peers can't offer), and opt-in keeps it zero-burden for everyone else. Because it'll be a *shipped* security surface (even if optional), we want to keep it airtight over the long run — so we'd really value your continued input as it evolves; realistically we'll want to check in with you from time to time as the module grows.

Here's the path to merge. One blocker, then a few honesty/scope items:

### Blocker — silent plaintext downgrade on recall
`persist_access_updates` (`src/memory/storage.rs:2029`) serializes with `crate::serialization::encode_sho(memory)` directly instead of `encode_memory(memory)`, so it **skips encryption**. It runs on the recall hot path (to bump `access_count` / `last_accessed` / importance), and since `deserialize_memory` transparently accepts plaintext, **every memory that's ever recalled has its on-disk ciphertext overwritten with plaintext** — silently, no error. One-line fix (`encode_sho` → `encode_memory`), but it needs a keystore-on round-trip test (store → recall → reopen the DB → assert the record is still `is_encrypted_record` and the raw bytes don't contain the plaintext). The current tests miss it because they run without a keystore.

### Before merge
1. The `persist_access_updates` fix + round-trip test above.
2. An **encryption-required tripwire** — a WARN/metric (or a strict env like `SHODH_REQUIRE_ENCRYPTED_READS`) when a plaintext record is read under an active keystore. That's the guard that would have caught #1, and it closes the plaintext-downgrade path for good.
3. **Scope, stated honestly** — encryption currently covers the primary `Memory` record only; facts, graph episodes/entities/edges, and **embeddings** stay plaintext (and embeddings invert closely enough to reconstruct content). Either extend coverage, or document the v2 scope precisely so "encrypted at rest" isn't oversold. Security-focused users will ask *exactly* what's encrypted.
4. Please **split `http_embedder.rs` into its own PR** — it's off-by-default and defensively guarded (nice work on the SSRF checks), so not a live leak, but an exfiltration-capable network client wants its own network/privacy review on its own terms rather than riding in under an encryption change.
5. **Ops hardening:** `shodh_keyctl` accepts secrets on argv (visible via `ps` / shell history) → env-only or a TTY prompt; no Windows ACL on the keystore file (our primary target OS) → owner-only; the `SHODH_KMS_WRAP_KEY` env path is a full unseal bypass readable via `/proc/<pid>/environ` → support a file/secret-store reference into `Zeroizing` and document that env delivery only defends disk theft, not host compromise; `rotate_dek` needs a persist-before-use invariant so a crash between rotate and save can't orphan records.

### Fast-follows (non-blocking)
HKDF-derive purpose-separated subkeys from the KEK (instead of using the raw KEK across GCM/HMAC/SHA-256), bind an AAD/encryption-context into the KMS wrap, return `Zeroizing` from `recovery_key_from_code` and `decode_32`, raise the KDF downgrade floor from 8 MiB to ≥19 MiB, authenticate the rollback-generation sentinel, and add `keystore.json` / `*.key` to `.gitignore`.

---

Would you be up for **driving the fixes**? It's your code and your domain, so it makes sense for you to lead — we'll review and merge. No rush on our end at all; we'd much rather it be airtight than fast, especially given who it's for. Genuinely excellent work — the core is in great shape, and we're keen to ship it.

varun29ankuS:
Thanks for the patience on this one — review verdict is positive on the crypto core and we'd like to land it, selectively.

Proposal: split this into (1) crypto core — keystore, rotation, at-rest encryption, opt-in and default-off; (2) KMS integration; (3) oblivious seam + PIR. We'd take (1) first, then (2); (3) needs its own design discussion.

Two requirements for (1): the encryption-requested-but-unavailable path must fail loud rather than silently writing plaintext (the storage.rs:2029 path in the current diff), with a contract test pinning that behavior; and a rebase on current main — the branch has drifted into conflicts and we want to review the real diff.

If you'd rather not split, let us know and we'll figure out an alternative — but smaller pieces will merge much faster.

Portll:
Taking the split you proposed. Part (1) is #555: crypto core only, ported hunk by hunk onto current main (three commits on `a724b86`), opt-in and default-off, with the `persist_access_updates` blocker fixed and pinned by the store → recall → reopen → raw-bytes contract test you described, the fail-loud contract test for keystore-present-but-unavailable, and the plaintext-read tripwire (WARN, plus `SHODH_REQUIRE_ENCRYPTED_READS` to refuse). KMS (part 2) and the oblivious seam + PIR (part 3) are not in it; the HTTP embedder will come on its own. Scope is stated in SECURITY.md as exactly the primary `Memory` record. Happy for this PR to be closed in favour of #555, or left open as the tracking issue for parts 2 and 3, whichever you prefer.

PR #335, opened 2026-06-24
feat(er): merge-quality harness + precision-first content-token merge gate

varun29ankuS:
## What

Two things, found by an isolated-A/B improvement loop over the entity-resolution and causal-chain harnesses:

1. **A merge-quality harness (`recall-eval --merge`)** — the first measurement of *cross-document entity-resolution* quality. The existing `--linking` harness measures query→graph linking, not whether mentions were correctly merged. The new harness plants known coref clusters (the same entity under surface variants, including a transliteration pair) plus distinct entities with tempting token collisions, resolves each surface through the graph, and scores the pairwise merge decisions: **merge precision, merge recall, and the false-merge rate** (distinct entities wrongly fused — the catastrophic metric).

2. **A content-token gate for the name-embedding merge (`SHODH_NAME_MERGE_TOKEN_GATE`, default off)** — requires a shared proper-noun token *on top of* the cosine threshold before merging two entities. Cosine stays required, so the gate can only **prevent** a merge, never create a false one.

## Why (the loop, in three iterations)

- **Iter 1** — `SHODH_EVAL_NAME_EMB` looked catastrophic: it collapsed the lineage harness (root-cause p@1 **0.883 → 0.000**). The lineage-only view said "reckless over-merger, revert."
- **Iter 2** — built the `--merge` harness, which **overturned that conclusion**: on real name variants, name-emb is a *good* resolver — merge recall **0.0 → 0.8**, **zero false merges**, and it correctly resolves the transliteration pair (Mohammed/Muhammad Ali) while keeping distinct collisions apart (Rajesh Verma, Mohammed Khan, Sunil Gupta). Its only failure is **templated names** ("the Vornak incident" vs "the Meslin incident"), where the embedding is dominated by the shared template and distinct events look identical.
- **Iter 3** — the token gate. Result (isolated A/B, control vs name-emb+gate):

| | control | name-emb + gate |
|---|---|---|
| merge recall | 0.0 | **0.8** |
| false-merge rate | 0.0 | **0.0** |
| lineage root-cause p@1 | 0.883 | **0.883** (held) |
| lineage direct-cause control | 1.0 | 1.0 |

A clean Pareto win: ER merge recall up, zero false merges, **and** the causal lineage no longer collapses.

## Notes

- `SHODH_NAME_MERGE_TOKEN_GATE` is **default off** — opt-in. It only affects the embedding-merge tier, which runs when name embeddings are populated.
- This is validated on the fast synthetic harnesses; a full-suite (LongMemEval) confirm is required before flipping any default.
- The `er-causal-loop.yml` workflow is the experiment driver for the loop; it can be archived alongside the other one-shot studies once this lands.

### DIFF
--- a/.github/workflows/er-causal-loop.yml
+++ b/.github/workflows/er-causal-loop.yml
@@ -1,0 +1,135 @@
+name: ER + causal improvement loop (isolated A/B)
+
+# The measurement engine for the entity-resolution + causal-chain improvement loop.
+# Runs TWO fast harnesses for control vs a candidate flag, in isolation, and reports
+# per-metric deltas so each iteration is "isolated and improved or reverted":
+#   * --linking  : query->graph entity-LINKING accuracy (precision = no false links =
+#                  the catastrophic-merge analog; recall = right seeds found; no-seed%).
+#   * --lineage  : planted causal-chain retrieval (root-cause p@1/recall vs direct control).
+# Iteration N changes the `flag` input (a candidate SHODH_* lever) and reads the deltas.
+# WIN rubric: ER  -> link recall UP and/or no-seed% DOWN with precision HELD (never trade
+#                    precision for recall — a wrong link is worse than a missing one).
+#             causal -> root-cause p@1/recall UP with the direct-cause control HELD.
+
+on:
+  workflow_dispatch:
+    inputs:
+      flag:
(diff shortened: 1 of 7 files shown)


varun29ankuS:
Closing — experimental with red CI; entity-resolution / merge-quality is now covered by the powerhouse CESI path.

PR #338, opened 2026-06-25
feat(causal): SHODH_CATENA — comprehensive causal sieve (root-cause P@1 0.58->1.00 on hard harness)

varun29ankuS:
## What

`SHODH_CATENA` — a comprehensive causal extraction sieve (CATENA sieves 1+2), replacing the 10 hand-picked causal cues with a real lexicon.

- **~130-term causative lexicon, classed by strength**: strong causative verbs (FrameNet "Causation" / VerbNet causal frames — *precipitate, engender, catalyze, instigate, foment, beget, spawn, kindle…*) → `Causes`; enabling verbs (*enable, facilitate, pave the way…*) + causal discourse markers (*because, owing to, in the wake of, stemmed from, by virtue of…*) → `Triggers`.
- **Comprehensive effect-first direction** — full inventory of passive causatives + consequence/origin connectives, so surface-inverted causation ("D was caused by C", "a consequence of C") gets the arrow right.
- **Negation precision sieve** — suppresses causation a sentence denies ("did not cause", "without triggering") to avoid false causal edges.

Closed lexicon, no parser. Default-OFF flag, checked before the generic relation cues.

## Measured — hard lineage harness (`SHODH_LINEAGE_HARD`, 3-hop / real names / inverted cues)

| | baseline (no CATENA) | **CATENA** |
|---|---|---|
| root-cause P@1 (+spreading) | 0.583 | **1.0000** |
| root-cause recall@10 | 0.667 | **1.0000** |
| `Causes` edges | 30 | **120** |

At the causal-spreading stage the backward origin-walk now reaches **every** root cause — the chains the base extractor missed (`followed directly from`, `a consequence of`) and the inverted-direction cases extract correctly-directed causal edges.

## Known follow-up (not in this PR)

The **full** pipeline drops to 0.52 (from 1.0 at +spreading) — the rerank/fusion layers displace the root cause the graph stage ranked #1. That's a fusion/ranking issue (the causal-origin boost not surviving rerank), separate from extraction. Next.

Bases on `feat/lineage-hard-harness` (#337, the measurement prerequisite). Sieve 3 (temporal precedence) + event-level extraction need the dependency parser — the apex program.


varun29ankuS:
Closing — superseded by the extraction-integration path (powerhouse GLiREL + OpenIE + CATENA → shodh-native increment), so a standalone SHODH_CATENA branch is redundant.

PR #358, opened 2026-07-02
Fix proactive context dropping explicit overlap matches

### LINKED ISSUE #283: proactive_context returns empty while recall finds exact memory at 0.95
## Summary

In practice, `proactive_context` has never surfaced useful memories for my workflows across projects/chats. I am reporting this specific reproducible case because it shows the pattern clearly: the same memory is returned by `/api/recall` at high confidence, but `/api/proactive_context` returns no results.

## Environment

- Shodh backend: `0.2.0`
- Package: `@shodh/memory-mcp@0.2.0`
- Backend health: `{"status":"healthy","version":"0.2.0"}`
- Index verification: healthy, `12491` memories, `12492` vectors, `0` orphaned

## Minimal reproduction

I stored a generic workflow preference memory. Sanitized example:

```text
Workflow preference: When the user asks to run an automated validation task, use the existing local runner path and avoid triggering duplicate remote automation runs. Do not create temporary branches or run extra remote jobs unless explicitly requested. Important keywords: automated validation, local runner, duplicate remote automation, workflow dispatch, temporary branches.
```

Then I queried recall with the same semantic context:

```bash
curl -sS -H "Content-Type: application/json" -H "X-API-Key: <key>" \
  -d '{"user_id":"claude-code","query":"run automated validation using the local runner without duplicate remote automation","limit":10,"mode":"hybrid"}' \
  http://127.0.0.1:3030/api/recall
```

Observed: recall returns the memory as the top result with score `0.95`.

Then I queried proactive_context with the same semantic context:

```bash
curl -sS -H "Content-Type: application/json" -H "X-API-Key: <key>" \
  -d '{"user_id":"claude-code","context":"run automated validation using the local runner without duplicate remote automation","max_results":10,"semantic_threshold":0.1}' \
  http://127.0.0.1:3030/api/proactive_context
```

Observed response:

```json
{"memories":[],"due_reminders":[],"context_reminders":[],"memory_count":0,"reminder_count":0,"todo_count":0}
```

I also reproduced the same behavior through the MCP tool: explicit `recall` finds the memory, but `proactive_context` returns no relevant memories.

## Expected behavior

`proactive_context` should surface the same highly relevant memory, or at least return it when `semantic_threshold` is lowered to `0.1` and `max_results` is increased.

## Why this matters

The MCP tool description says `proactive_context` should be called with every user message to surface relevant memories automatically. If it returns empty even when `recall` finds an exact relevant memory at `0.95`, then it cannot be relied on as a proactive safety/context mechanism. For now I have to use explicit `recall(...)` for important project instructions.

GTanger:
## Summary
- restrict proactive_context anti-echo filtering to auto-ingested memories only
- keep explicit user-saved memories eligible even when they strongly overlap the current context
- apply semantic_threshold after proactive scores are normalized to the public 0.0-0.95 score scale
- add regression coverage for both the recall/proactive_context divergence and the threshold=0 vs tiny-positive-threshold behavior reported in #283

Fixes #283.

## Tests
- LD_LIBRARY_PATH=/home/tanger/.cache/shodh-memory/onnxruntime${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH} cargo test --test pipeline_integration_tests proactive_context_threshold_uses_normalized_score_scale -- --test-threads=1 --nocapture
- LD_LIBRARY_PATH=/home/tanger/.cache/shodh-memory/onnxruntime${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH} cargo test --test pipeline_integration_tests proactive_context -- --test-threads=1 --nocapture
- LD_LIBRARY_PATH=/home/tanger/.cache/shodh-memory/onnxruntime${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH} cargo test --test handler_pipeline_tests proactive_context -- --test-threads=1 --nocapture
- cargo fmt --all -- --check

### DIFF
--- a/src/handlers/recall.rs
+++ b/src/handlers/recall.rs
@@ -2054,10 +2054,15 @@ pub async fn proactive_context(
                     if alpha_count < 10 {
                         return false;
                     }
-                    // Anti-echo: skip memories that are just our own context echoed back
-                    // (auto-ingest stores context, which then gets retrieved for itself)
+                    // Anti-echo: skip auto-ingested memories that are just our own context
+                    // echoed back. Explicit memories can legitimately overlap the current
+                    // context and must still surface through proactive_context.
                     // Uses inline comparison to avoid allocating a HashSet<String> per candidate
-                    if !query_words.is_empty() {
+                    let is_auto_echo_candidate = m.experience.tags.iter().any(|tag| {
+                        tag.eq_ignore_ascii_case("auto-captured")
+                            || tag.eq_ignore_ascii_case("assistant-response")
+                    });
+                    if is_auto_echo_candidate && !query_words.is_empty() {
                         let mut mem_word_count = 0usize;
                         let mut overlap = 0usize;
                         for w in content.split_whitespace() {
--- a/src/handlers/recall.rs
+++ b/src/handlers/recall.rs
@@ -2291,20 +2291,23 @@ pub async fn proactive_context(
                 enriched.sort_by(|a, b| b.1.total_cmp(&a.1));
             }
 
-            // Drop results below minimum absolute score — don't pad with irrelevant filler
-            // Also drop results that are < 30% of the top score (too weak relative to best)
+            // Normalize scores before applying the public threshold. Raw pipeline scores are
+            // often tiny after proactive recency decay, while the API exposes scores on the
+            // same relative 0.0-0.95 scale as recall().
             let top_score = enriched.first().map(|(_, s, _)| *s).unwrap_or(0.0);
-            let abs_min = semantic_threshold;
-            let relative_min = top_score * 0.30;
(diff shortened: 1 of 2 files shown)


varun29ankuS:
Excellent work — verified both root causes against the pipeline before merging: the anti-echo overlap filter was dropping exactly the memories proactive_context exists to surface, and applying semantic_threshold to raw scores made anything older than ~2 days mathematically unreachable at sane thresholds (the proactive recency differential is exp(-0.02·h)). Bonus you didn't even claim: the reorder also fixes a units mismatch in the habituation gate, which was comparing normalized scores against a raw threshold. Regression tests included, tags match our ingest conventions exactly. Merging — thank you.

PR #361, opened 2026-07-03
docs: add direct server systemd guide

GTanger:
## Summary
- Add an optional Linux systemd guide for running `shodh server` as a supervised long-lived HTTP server.
- Show how MCP clients can connect through the native `shodh serve` bridge and how REST clients can call the same server.
- Document `SHODH_USER_ID`/`user_id` as a stable memory namespace, not an API key or secret, using generic example values.

## Motivation
This documents a direct-server deployment shape where the Rust HTTP server lifecycle is managed outside any one MCP client or npm wrapper. It gives users a clear path for health checks, journal logs, restarts, and multiple local clients sharing the same backend.

This PR does not change npm package or wrapper behavior.

## Validation
- `git diff --cached --check`
- Docs-only change; no Rust tests were run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,14 @@ claude mcp add shodh-memory -- npx -y @shodh/memory-mcp
 </details>
 
 <details>
+<summary>Direct server mode (systemd / MCP / REST)</summary>
+
+For Linux users who want the Rust HTTP server supervised separately from MCP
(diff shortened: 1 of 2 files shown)


varun29ankuS:
Merged with thanks — verified every documented command and env var against the CLI; all real. Good to have the direct-server deployment shape written down.

Issue #90, opened 2026-03-04
Memory leak: shodh-memory-server grows to 2GB+ RAM over 24 hours


korboybeats:
## Description

The `shodh-memory-server` binary (v0.1.74) exhibits a significant memory leak when running as a persistent backend across multiple Claude Code sessions.

## Observed Behavior

- **Fresh start**: ~23MB RSS
- **After ~25 hours**: ~2GB RSS (2,046,876 KB)
- **System impact**: On a 7.7GB RAM machine, the server consumed 25% of total memory
- **Symptom**: MCP tool calls start failing with "Failed after 3 attempts: This operation was aborted" (timeouts)

## Environment

- **OS**: Linux 6.6.87.2-microsoft-standard-WSL2 (Ubuntu on WSL2)
- **RAM**: 7.7GB total
- **Server version**: 0.1.74
- **MCP package**: `@shodh/memory-mcp` via `npm exec`
- **Concurrent sessions**: 4 Claude Code sessions, each with their own MCP frontend process, all hitting the single shared backend on port 3030

## Reproduction

1. Start `shodh-memory-server`
2. Use Claude Code normally across multiple sessions over 24+ hours
3. Each session calls `proactive_context` on every user message (with `auto_ingest: true`)
4. Monitor RSS with `ps -p <pid> -o rss`

## Workaround

Kill and restart the backend server. Memory drops back to ~23MB immediately.

## Possible Causes

- Unbounded caching of embeddings or NER inference results
- In-memory graph/node accumulation without eviction
- TinyBERT inference tensor allocations not being freed
- Feedback loop state accumulation across many `proactive_context` calls

varun29ankuS:
This is a known RocksDB upstream issue — we have a PR with them. Not actionable on our side until that lands.

GTanger:
Adding a current-source observation in case it is useful for tracking the Linux side of this issue. I understand from the maintainer notes above that the remaining root cause is largely RocksDB/allocator behavior, so this is intended as reproduction evidence rather than a claim that I have a server-side fix.

Environment:

- OS: Linux desktop, systemd user service
- shodh runtime: source-built from upstream `main`
  `4dc8bdd5cf71e988c56c8e160590ff0d54db90f4`
- tag/package line still reports `0.2.0`
- `shodh health`: healthy, `write_failures=0`, `pending_retries=0`
- data scale at observation time: about `3013` memories, `12206` graph nodes, and about `250k` graph edges

Observed:

[...]

PR #362, opened 2026-07-03
fix(metrics): avoid loading cold users during scrape

GTanger:
## Description

Make `/metrics` aggregate per-user tier/vector/estimated-heap gauges from currently cached users only.

Previously the endpoint iterated over `list_users().take(100)` and called `get_user_memory()` for each entry. `list_users()` returns filesystem user directories, so a metrics scrape could open cold RocksDB-backed user instances and increase memory/FD pressure while observing memory.

This mirrors the existing `/health/index` aggregate path, which already uses `list_cached_users()` to keep the endpoint side-effect-light.

Refs #90.

## Testing

- `cargo test handlers::health::tests::metrics_endpoint_does_not_load_cold_disk_users --lib -- --exact --test-threads=1`
- `cargo test --lib -- --test-threads=1`
- `cargo clippy --lib --bins`
- `git diff --check`

## Notes

`cargo test --lib -- --test-threads=1` passed with 774 tests, 4 ignored.

`cargo clippy --lib --bins` exits 0, with existing warnings outside this diff. `cargo clippy --lib --bins -- -D warnings` is not currently a clean baseline in this checkout due to unrelated existing warnings, so this PR intentionally keeps those cleanup items out of the metrics behavior fix.

### DIFF
--- a/src/handlers/health.rs
+++ b/src/handlers/health.rs
@@ -170,13 +170,15 @@ pub async fn metrics_endpoint(State(state): State<AppState>) -> Result<String, S
     let users_in_cache = state.users_in_cache();
     metrics::ACTIVE_USERS.set(users_in_cache as i64);
 
-    // Aggregate metrics across all users
+    // Aggregate metrics across currently cached users only.
+    // list_users() returns filesystem directories; get_user_memory() would open
+    // RocksDB for cold users and make a metrics scrape increase memory/FD usage.
     let (mut total_working, mut total_session, mut total_longterm, mut total_heap) =
         (0i64, 0i64, 0i64, 0i64);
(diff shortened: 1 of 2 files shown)


varun29ankuS:
Verified and merging — sharp find. Confirmed against main: the scrape really did open cold RocksDB instances via list_users().take(100), and /health/index already established the list_cached_users() pattern with the same rationale, so this brings /metrics in line. The observer was creating the memory pressure it observes — a Prometheus poller would progressively fault in every user directory on disk, which makes this a plausible real contributor to #90. Semantic note for the record: tier gauges now aggregate cached users only (the semantics /health/index already chose). Thank you — three for three today.

PR #371, opened 2026-07-03
ci: add optional PMI gate threshold sweeps

GTanger:
## Summary

Adds an opt-in `gate_min_sweep` workflow input to the existing PMI gate ablation workflows:

- `.github/workflows/pmi-gate-ablation.yml`
- `.github/workflows/pmi-gate-longmemeval.yml`

When left blank, the workflows behave exactly as they do today. When provided as comma-separated values, the workflow runs extra gated arms and appends a compact table with recall and edge-provenance counters for each `SHODH_GRAPH_PMI_GATE_MIN` value.

Also documents the PMI gate interpretation boundary: it only gates generic edge births, does not shrink existing graphs retroactively, and its effect can depend heavily on corpus shape.

## Why

A single `gate_min=0.0` A/B run can be inconclusive on corpora that are already pre-processed, tag/entity-projected, or otherwise not raw dialogue imports. The optional sweep gives maintainers a cheap manual diagnostic before changing defaults or interpreting a local graph profile.

## Verification

- Parsed both workflow YAML files with PyYAML.
- Parsed all embedded Python heredocs with `ast.parse`.
- `git diff --check`
- `cargo fmt --all -- --check`
- `cargo test --no-run`
- `cargo test --lib --bins` (`790` lib tests passed, `4` ignored; CLI `2` passed)

I did not run the full workflow bodies locally because they require the benchmark/model download path.

### DIFF
--- a/.github/workflows/pmi-gate-ablation.yml
+++ b/.github/workflows/pmi-gate-ablation.yml
@@ -23,6 +23,10 @@ on:
         description: 'SHODH_GRAPH_PMI_GATE_MIN (PMI floor; 0.0 = PPMI). Higher prunes more.'
         required: false
         default: '0.0'
+      gate_min_sweep:
+        description: 'Optional comma-separated SHODH_GRAPH_PMI_GATE_MIN values for extra gated arms.'
+        required: false
+        default: ''
       max_cases:
         description: 'SHODH_MAX_CASES cap (blank = all)'
         required: false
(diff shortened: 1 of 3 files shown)


varun29ankuS (changes requested):
Thanks — we checked the mechanics against the source and they line up: log field names, report JSON shape, opt-in guard, pinned env vars, and the doc note all match the implementation.

Two findings before merge:

1. In `pmi-gate-longmemeval.yml` the sweep arms run sequentially after A and B under the 360-min cap. One arm at limit=50 is ~1.5h, so two sweep values already brush the cap and three gets killed after burning the whole budget — we lost a run to exactly this last week (#372). Suggest documenting a max sweep size, guarding it in `parse_gates`, or going matrix-parallel.

2. Fragment-blocked pairs *are* stored (they get `CoOccurs` and proceed to edge creation), so the LongMemEval version's `stored_total` undercounts by omitting `fragment_blocked` — the ablation version has it right.

Both small; happy to merge once addressed.

GTanger:
Addressed in `ffaa2a7`:

- Capped the LongMemEval `gate_min_sweep` path to one extra gated arm per dispatch and updated the input/help text. That keeps the default `limit=50` run away from the 360-minute timeout; more values should be run as separate dispatches.
- Fixed the LongMemEval sweep edge accounting to include `fragment_blocked` in `stored_total`, matching the ablation workflow's accounting.
- Tightened the doc note with the issue #90 clarification: PMI gate is a hub-noise filter, so hub-shaped corpora can shrink sharply while clique-shaped corpora with many rare pairs may not.

Rechecked locally:

- workflow YAML parse
- embedded Python `ast.parse`
- `git diff --check`
- `cargo fmt --all -- --check`

GTanger:
Synced the branch with current `main` in `1a44119` after #372 landed, so the PR is no longer behind upstream. This is a merge-only update; the requested-change fix remains `ffaa2a7`.

Rechecked after the sync:

- workflow YAML parse
- embedded Python heredocs via `ast.parse`
- `git diff --check`
- `cargo fmt --all -- --check`

CI has restarted on the updated branch.

varun29ankuS (approved):
Verified ffaa2a7 against both findings: the one-arm cap is enforced in parse_gates (not just documented) and the fragment_blocked accounting now matches the ablation workflow. The hub-vs-clique doc framing is a nice addition. Thanks for the fast turnaround.

PR #374, opened 2026-07-04
feat(config): tune RocksDB block cache capacity

GTanger:
## Why
#90 decomposed the steady-state RSS plateau into a warm shared RocksDB block cache, ONNX arenas, and baseline process memory. The shared block cache is intentionally bounded, but today its 256MiB ceiling is hard-coded. That is a good default for general workloads, while edge/local deployments may want to trade cache hit rate for lower steady-state RSS.

While tightening that observability path, I also found a cached-user race: some health/metrics/telemetry aggregations listed cached user IDs and then called `get_user_memory()`. If a user was evicted between those two steps, the observation path could reopen a cold RocksDB instance from disk, which works against the "diagnostics must not create the pressure they observe" guard.

## What
- Add `SHODH_ROCKSDB_BLOCK_CACHE_MB` as an optional MiB override for the shared RocksDB block cache.
- Keep the existing 256MiB default when the env var is unset, zero, invalid, or overflowing.
- Store the actual configured capacity on the manager so `/health.rocksdb_memory.shared_block_cache_capacity_bytes` reports the real ceiling.
- Expose `shodh_rocksdb_block_cache_capacity_bytes` in Prometheus metrics so operators can alert on cache usage relative to the configured ceiling.
- Add a cached memory snapshot helper and use it in health, metrics, write-failure aggregation, flush/save loops, and telemetry paths that must not reopen cold users.
- Document the env var in README production config and CLI env help.

## Tested
- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo check`
- `cargo test rocksdb_shared_cache_capacity`
- `cargo test metrics_endpoint`
- `cargo test cached_user_memories_snapshot_does_not_reopen_evicted_users`
- `cargo test rocksdb_diagnostics_do_not_load_cold_disk_users`
- `cargo test metrics_endpoint_does_not_load_cold_disk_users`
- `cargo test --lib --bins`
- `cargo clippy --all-targets -- -W clippy::all`

Note: I did not run `cargo build` because this repo's `CLAUDE.md` says not to run `cargo build` or `cargo run` unless explicitly asked.

Refs #90

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,6 +359,7 @@ SHODH_PORT=3030                   # Port (default: 3030)
 SHODH_MEMORY_PATH=/var/lib/shodh  # Data directory
 SHODH_REQUEST_TIMEOUT=60          # Request timeout in seconds
 SHODH_MAX_CONCURRENT=200          # Max concurrent requests
+SHODH_ROCKSDB_BLOCK_CACHE_MB=256  # Shared RocksDB block cache (MiB)
 SHODH_CORS_ORIGINS=https://app.example.com
 ```
 </details>
--- a/src/config.rs
+++ b/src/config.rs
@@ -595,6 +595,7 @@ pub fn print_env_help() {
     println!("  SHODH_RATE_BURST       - Burst size (default: 8000)");
     println!("  SHODH_MAX_CONCURRENT   - Max concurrent requests (default: 200)");
     println!("  SHODH_REQUEST_TIMEOUT  - Request timeout in seconds (default: 60)");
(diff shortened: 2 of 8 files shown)


varun29ankuS (approved):
Verified against source: the capacity is stored on the manager so /health and Prometheus report the real ceiling, the parse falls back safely on zero/invalid/overflow, and the snapshot helper fixes a genuine reopen race in the #365 diagnostics path (list-then-get could reopen an evicted user — good catch, that was ours). Ran the diagnostics paths locally on Windows as well.

PR #375, opened 2026-07-04
fix(mcp): resolve displayed version from package metadata

### LINKED ISSUE #360: Server version seems to be hardcoded to version "0.1.90"
Hello, basically issue is as written in the title. I didn't dig much into this, but:

```
curl -X GET http://localhost:3030/health
{ "status":"healthy","version":"0.2.0", ... }
```

while at the exact same time logs shows:

```
[Stream] WebSocket connected to ws://127.0.0.1:3030/api/stream
...
[shodh-memory] Backend server started successfully
Shodh-Memory MCP server v0.1.90 running
Connecting to: http://127.0.0.1:3030
```

and `grep`ping for string "0.1.90" returns 2 matches:

```
 ~ % grep -c "0.1.90" [...]/node_modules/@shodh/memory-mcp/dist/index.js
2
```

Not a huge issue and I didn't dig any deeper, but this caused me to scratch my head quite a bit today :)

GTanger:
## Why

Issue #360 reports the MCP wrapper logging `Shodh-Memory MCP server v0.1.90 running` while the backend `/health` endpoint reports `0.2.0`. The source already wants the wrapper version to come from package metadata, so this makes that path regression-tested and removes the remaining hard-coded `v0.1.90` from the MCP README.

## What

- Move MCP package-version resolution into a small `version.ts` helper.
- Keep both published `dist/index.js -> ../package.json` and development `index.ts -> ./package.json` layouts supported.
- Add unit coverage for both layouts and the `unknown` fallback.
- Remove the hard-coded `v0.1.90` badge text from `mcp-server/README.md`; the npm badge remains the source of displayed package version.

## Tested

- `bun -e ...` resolver smoke for dist layout, dev layout, and fallback
- `bun build index.ts --outdir "$tmpdir" --target node --external @modelcontextprotocol/sdk`
- `rg -n "0\.1\.90" mcp-server --glob "!node_modules"` (no matches)
- `git diff --check`

Not run locally: full `vitest` suite, because this checkout has no MCP server dependencies installed and both `bun install` / `npm exec --yes vitest` stalled while resolving packages.

Fixes #360

### DIFF
--- a/mcp-server/README.md
+++ b/mcp-server/README.md
@@ -5,7 +5,7 @@
 <h1 align="center">Shodh-Memory MCP Server</h1>
 
 <p align="center">
-  <strong>v0.1.90</strong> | Persistent cognitive memory for AI agents
+  Persistent cognitive memory for AI agents
 </p>
 
 <p align="center">
--- a/mcp-server/index.ts
+++ b/mcp-server/index.ts
@@ -31,29 +31,12 @@ import { fileURLToPath } from "url";
 import { nextReconnectDelay, serializeAndValidateBody, shouldWarnInsecureApiUrl } from "./security-utils";
 import { stripSystemNoise, getContent as _getContent, getType as _getType, formatSurfacedMemories as _formatSurfacedMemories, formatToolCallContent } from "./string-utils";
 import { TokenTracker } from "./token-tracking";
+import { resolvePackageVersion } from "./version";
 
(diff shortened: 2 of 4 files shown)


varun29ankuS (approved):
Verified: resolver behavior is unchanged, now regression-tested with an injectable reader, and the stale v0.1.90 badge is gone. One note on our side: CI doesn't currently run the mcp-server vitest suite, so your new tests won't execute until we wire that in — tracked as a follow-up on our end.

PR #376, opened 2026-07-04
feat(graph): add generic pair budget

GTanger:
## Description
Add an optional `SHODH_GRAPH_GENERIC_PAIR_BUDGET` throttle for generic graph edge births.

This is a follow-up to the #90 clique-shaped corpus discussion: PMI gating filters hub noise, but rare-per-memory cliques can still birth many weak `CoOccurs` / `RelatedTo` edges. This PR gives maintainers a default-off measurement/control lever without changing existing behavior.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Implementation
- Adds `ServerConfig::max_generic_edges_per_memory`, populated from `SHODH_GRAPH_GENERIC_PAIR_BUDGET`.
- Applies the budget after the PMI gate and only to non-fragment generic `CoOccurs` / `RelatedTo` edge births.
- Preserves typed/cue/semantic/label edges and fragment-safe bridges even when the budget is `0`.
- Adds provenance counters for kept/gated generic-budget edges.
- Documents the knob in README and the Hebbian/PMI architecture notes.

## Testing
- [x] `cargo fmt --all -- --check`
- [x] `git diff --check`
- [x] `cargo test generic_pair_budget --lib -- --test-threads=1`
- [x] `cargo test --test handler_tests graph_generic_pair_budget -- --test-threads=1`
- [x] `cargo check --all-targets`
- [x] `cargo clippy --all-targets -- -W clippy::all` (exit 0; existing upstream warnings remain)

Not run locally: `cargo build` / `cargo run`.

## Notes
This does not compact existing dense graphs retroactively; it only affects new graph construction when the budget is configured.

Relates to #90.


varun29ankuS (changes requested):
The mechanism checks out: the budget correctly runs after the PMI gate, keeps pairs in genuine salience/name order (we verified `all_entities` is sorted before the pair loop), exempts typed edges and fragment bridges, and the counters slot cleanly into the `edge typing provenance` line. Tests also pass locally on Windows.

One change before merge, and it goes to the PR's stated purpose: the knob is routed through `ServerConfig::from_env`, but the recall harness builds its manager with `ServerConfig::default()` (`runner.rs:868`) — so `SHODH_GRAPH_GENERIC_PAIR_BUDGET` is invisible to every recall A/B workflow. All the sibling knobs in that function (PMI gate, extracted predicates, learned pairs) read their env var directly at the use site, which is what makes them measurable. Suggest reading the env var directly there (keeping the ServerConfig field for server use is fine), so we can actually run the recall-safety A/B this lever needs before anyone turns it on.

Minor: the README example lists `SHODH_GRAPH_GENERIC_PAIR_BUDGET=20` inline with standard config, which reads as a default — worth marking it optional/off-by-default there.

varun29ankuS:
Superseded by the clean-KG rebuild in #389 — strengthen-not-create, the PMI gate (default-on), and the typed-only gate now bound generic co-occurrence at mint, which is what the pair budget was for. Closing rather than rebasing a redundant knob onto the rewritten minting path.

PR #395, opened 2026-07-14
feat: add authenticated local IPC transport

yanchenko:
## Summary
- add a versioned, bounded local IPC server over Unix-domain sockets and Windows named pipes
- reuse the canonical Axum router so REST, Rust MCP, TypeScript MCP, and local integrations share one operation implementation
- require API-key authentication for protected routes and enforce owner-scoped endpoint permissions
- route Rust MCP through IPC and let TypeScript MCP select IPC explicitly while preserving its existing HTTP compatibility mode
- keep streaming on its existing WebSocket path as an explicit opt-in when IPC is active

## Verification
- independent architecture and security review: approved
- independent Windows saturation review: approved, bounded at four helpers with no leaked processes or native operations
- Rust format, checks, focused IPC tests, Clippy, and the pre-rebase 839-test library suite passed
- TypeScript: 107 tests, static check, and production bundle passed after the final queue fix
- post-rebase full library test was intentionally stopped at the requester's direction so CI can complete it


yanchenko:
Recreating this as a fork-local draft pull request.

varun29ankuS:
Caught this before the re-draft — direction is right: one canonical router across REST/MCP/IPC is exactly how this should compose, and a no-open-TCP-port local transport fits the local-first posture of the project. Looking forward to the draft when it's ready. Heads-up: main moved a lot this week (new entity typer landed, `state.rs` churned heavily) — ping me if the rebase fights you.

PR #399, opened 2026-07-14
feat: add authenticated local IPC transport

yanchenko:
## Description

Add an authenticated local IPC transport for Shodh-Memory so local MCP integrations can avoid loopback HTTP while continuing to use the canonical Axum route table.

The server listens on an owner-scoped Unix-domain socket or per-user Windows named pipe. The native Rust MCP client prefers IPC with HTTP fallback; the TypeScript MCP client selects IPC explicitly through `SHODH_IPC_ENDPOINT`. Existing REST clients remain compatible.

### What changed

- Added bounded, versioned newline-delimited JSON request and response framing.
- Added Unix socket ownership, permission, stale-endpoint, and cleanup safeguards.
- Added Windows named-pipe security restricted to the current user and LocalSystem.
- Preserved API-key authentication, route policy, deadlines, correlation, and shutdown draining.
- Added native Rust and TypeScript IPC clients with focused regression coverage.
- Made recall asset downloads work for fork pull requests.
- Documented configuration, platform endpoints, security boundaries, fallback behavior, service verification, and known limitations.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

- [x] Unit tests pass: 872 passed, 0 failed, 4 ignored
- [x] Integration tests pass
- [x] Manual Windows named-pipe testing completed
- [x] TypeScript MCP tests pass: 107 passed
- [x] Formatting and Clippy pass
- [x] Release builds pass on Ubuntu, macOS, and Windows
- [x] Recall smoke gate passes within tolerance

## Checklist

- [x] Code follows style guidelines
- [x] Self-review completed
- [x] Comments added for complex code
- [x] Documentation updated
- [x] No new warnings generated
- [x] Tests added and updated
- [x] All required tests pass

## Security and compatibility

Protected IPC requests carry the configured API key. Exact unauthenticated access is limited to `GET /health`; sensitive values are redacted from error paths. Unix endpoints are owner-scoped, and Windows pipes use a protected per-user access-control list.

The IPC listener is enabled by default and can be disabled with `SHODH_IPC_ENABLED=false`. Native Rust clients fall back to HTTP when IPC is unavailable. TypeScript clients use HTTP unless `SHODH_IPC_ENDPOINT` is explicitly configured.

## Rollback

Disable the listener with `SHODH_IPC_ENABLED=false`, unset `SHODH_IPC_ENDPOINT` for TypeScript clients, or revert the documentation and feature commits.


varun29ankuS (changes requested):
Thanks for bringing this back as a clean PR — the architecture is right (reuse the canonical router instead of forking the route table) and much of the implementation is careful. I verified **auth-parity holds end-to-end**: IPC dispatches through the same `validate_api_key` before method/policy checks, and the IPC-reachable set is strictly ⊆ HTTP's protected routes — nothing authed on HTTP becomes unauthed on IPC. Constant-time key compare, `reject_remote_clients`, `first_pipe_instance`, frames bounded before allocation, per-connection panic isolation, no body/secret logging, and no runtime IPC→HTTP downgrade all check out. The Rust client correctly gets tokio's `SECURITY_IDENTIFICATION` SQOS.

Because shodh targets local-first / sovereign deployments, I reviewed against an adversarial *same-host* model (a compromised or sandboxed co-resident process, multi-agent setups). Under that lens there's a must-fix set before this ships default-on:

**Critical — Windows client impersonation.** `mcp-server/ipc-client.ts:22` opens the pipe with `fs.openSync(..., O_RDWR)`, which reaches `CreateFileW` *without* `SECURITY_SQOS_PRESENT`, so the pipe defaults to `SecurityImpersonation`. Paired with pipe-name squatting, a co-resident process that pre-creates `\.\pipe\shodh-memory-<SID>` becomes the server, and once the victim client connects it can `ImpersonateNamedPipeClient()` to obtain a `SecurityImpersonation` token of the victim — token theft, cross-integrity if the client runs elevated. The Rust `IpcClient` is unaffected (tokio sets `SECURITY_IDENTIFICATION`). Fix: route the Node client through the Rust client, or open the pipe with `SECURITY_SQOS_PRESENT | SECURITY_IDENTIFICATION` via a small shim.

**High — squatting forces a silent downgrade + key harvest.** With `first_pipe_instance(true)`, an attacker who squats the name during a boot/restart window makes the real server's bind fail; `server.rs:313-317` then degrades to HTTP-only and only logs it (the comment at :324 says it "must never degrade silently" — but it does), and the client hands the full-authority key to the squatter on the very first `/health` probe, since the client always serializes the key. Fix: (a) verify the endpoint owner before sending the key — `GetNamedPipeServerProcessId` + SID on Windows, `getpeereid`/`SO_PEERCRED` on Unix; (b) send empty `auth` on the health probe; (c) a `SHODH_IPC_REQUIRED=1` fail-closed mode (server bind fatal, client refuses HTTP fallback).

**High — body `user_id` is trusted for tenant selection.** Handlers call `get_user_memory(&req.user_id)` after only format-validating it, so any holder of the shared key can read/write another tenant by changing `user_id`. It's inherited from the HTTP model, but IPC is where per-peer identity becomes available — derive the effective tenant from the authenticated/peer identity and reject a mismatched body `user_id`.

**Medium** — the 8 MiB `body: Value` is deserialized *before* the auth check and a connection holds one of 200 slots for the full 60s timeout pre-auth, so one peer can pin all slots / amplify memory (`body: Box<RawValue>` + validate `auth` first + short pre-auth deadline + per-peer cap); no server-side peer-credential assertion; transport selection is silent and HTTP stays co-bound (a `SHODH_IPC_REQUIRED` switch covers this too).

**Low** — narrow TOCTOU on the Unix parent `chmod 0700` if the grandparent is writable (use `O_DIRECTORY|O_NOFOLLOW` + `fchmod`); the constant-time compare is correct but consider `subtle::ConstantTimeEq` so it stays that way across compiler upgrades; the LocalSystem-service case widens the DACL to all-SYSTEM.

Separately: the `recall.yml` asset-source change is fine and I confirmed it's safe (under `pull_request`, read-only token, no secrets), but it's unrelated to the transport — could you split it into its own PR? Optional while there: mirror the runtime's SHA-256 checksum into the CI download so both share one integrity anchor.

Findings 1 and 2 share a root cause — the client trusts the endpoint name and hands over both its token and its key before authenticating the server — so I'd want those two closed before merge. Strong foundation otherwise. If it's useful I'm glad to sketch the client-side server-owner-verification + SQOS change in more detail (it's the linchpin for both) — but it's your PR, so however you'd prefer to drive it. Happy to go deeper on any of these.

yanchenko:
Addressed in `b1d2e46`.

Key changes:

- Replaced the production Node named-pipe open with the bundled Rust helper. It uses `SECURITY_IDENTIFICATION` SQOS and verifies the server process SID before writing.
- Upgraded IPC to protocol v2. Health probes send no reusable key; challenge/response and request HMACs authenticate the endpoint and bind the request body.
- Added Unix peer-UID checks and Windows client/server process-account checks.
- Added `SHODH_IPC_REQUIRED` fail-closed behavior for server binding and client transport selection.
- Kept request bodies as `RawValue` until authentication, with a separate 16-connection/two-second pre-authentication limit.
- Fixed the Unix directory-permission TOCTOU by opening with `O_DIRECTORY | O_NOFOLLOW` and tightening permissions through the handle.
- Replaced the handwritten key comparison with `subtle::ConstantTimeEq`.
- Removed the unrelated `recall.yml` change from this PR.

For `user_id`, I did not derive application identifiers such as `claude-code` from an OS UID/SID because there is no sound mapping between them. API keys remain documented as server-wide, full-authority credentials across logical namespaces for both HTTP and IPC. Mutually untrusted tenants must use separate server/key instances; introducing tenant-scoped keys should be an auth-model change shared by both transports rather than an IPC-only identity rewrite.

Verification:

- `cargo check --all-targets --locked`
- `cargo fmt --all --check`
- `cargo test local_ipc --lib --locked` — 12 passed
- TypeScript test suite — 108 passed
- TypeScript production bundle completed successfully

varun29ankuS:
Re-reviewed `b1d2e46` adversarially (two passes, same-host threat model) — this is excellent and it over-delivers: the HMAC challenge-response mutual-auth handshake is exactly the right primitive, and it closes the original impersonation/key-harvest cleanly (key never on the wire, client authenticates the server before signing, body+route bound, v1 downgrade gone). All seven items are addressed, auth-parity and memory-safety hold, and the `user_id`/tenant deferral is the correct call — that belongs in a shared HTTP+IPC auth-model change, not here.

We're going to merge this and take the last-mile defence-hardening in-house rather than hold your PR through another round:
1. HKDF-derive an IPC-only subkey so the probe MAC never operates on the raw HTTP bearer key (today a same-user peer can treat the probe as a chosen-challenge MAC oracle over a weak key).
2. Request anti-replay (seen-id + short timestamp window, or a per-connection server challenge folded into the request MAC) — the per-process nonce alone leaves captured request frames replayable.
3. `SHODH_IPC_REQUIRED` as the sovereign-profile default, plus a real cross-implementation (Rust server ↔ TS client) auth test covering tamper/replay/stale-nonce.

None of these are exploitable outside the same-user trust domain, so they're follow-ups rather than blockers. Genuinely strong work — thank you for the rigor.

varun29ankuS (approved):
Approving — all 7 review findings closed/deferred-correctly; three defence-hardening items tracked as in-house follow-ups. Strong work.

Issue #12, opened 2026-01-22
Improve user experience and ease of installation


varun29ankuS:
## Problem

New users face significant friction when trying to use shodh-memory. Issue #11 highlighted that even basic TUI usage requires opening a support issue.

### Installation
- [ ] TUI not included in releases — requires Rust toolchain to build from source
- [ ] ONNX Runtime dependency adds complexity (even though it's bundled in releases)
- [ ] Multiple binaries to understand (`shodh-memory-server`, `shodh`, wrapper scripts)
- [ ] No package manager support (homebrew, winget, apt, etc.)

### Configuration
- [ ] Environment variables not discoverable — users must read code or docs
- [ ] No guided setup or configuration wizard
- [ ] Default API key (`sk-shodh-dev-local-testing-key`) looks like a real secret, causes confusion
- [ ] TUI requires manual `SHODH_SERVER_URL` if server is on non-default port

### Documentation
- [ ] README focuses heavily on MCP integration, TUI is barely mentioned
- [ ] No single "getting started" guide covering all components
- [ ] No troubleshooting section for common issues
- [ ] Architecture diagram would help users understand components

### Runtime
- [ ] TUI shows cryptic errors when server isn't running
- [ ] No health check or status command
- [ ] No way to verify installation is working

### Phase 1: Quick Wins
1. **Bundle TUI in releases** — add `shodh-tui` binary to release artifacts
2. **Add `--help` examples** — show common usage patterns in CLI help
3. **Better error messages** — "Server not running at localhost:3030. Start with: shodh-memory"
4. **Update README** — add TUI section with screenshots and instructions

### Phase 2: Unified CLI
Consolidate into single `shodh` binary with subcommands:
```
shodh server      # Start HTTP API server
shodh tui         # Launch TUI dashboard  
shodh mcp         # Run as MCP server (stdio)
shodh status      # Check if server is running
shodh init        # Interactive first-time setup
shodh doctor      # Diagnose common issues
```

### Phase 3: Package Managers
- [ ] Homebrew formula (macOS/Linux)
- [ ] Winget manifest (Windows)
- [ ] APT repository (Debian/Ubuntu)
- [ ] AUR package (Arch)

### Phase 4: Zero-Config Experience
- TUI auto-discovers local server (try common ports)
- Server auto-generates API key on first run, displays it once
- `shodh init` creates config file with sensible defaults
- Single install command that sets up everything

## Success Criteria

A new user should be able to:
1. Install with one command
[...]

PR #428, opened 2026-08-06
Fix the first five minutes: shared API key, backend lifecycle, hook install

varun29ankuS:
Fixes the four defects that make a fresh install fail — three of them silently. Relates to #12.

On a clean machine today: the shim generates an API key it never writes down, so hooks cannot learn it and every capture call 401s while the user believes memory is working. Closing one client kills the backend out from under the others. And `setup-hooks` writes a command that does not execute at all for users whose home directory contains a space.

## Changes

| Commit | Defect | Failure mode removed |
|---|---|---|
| `0481a926` | T15 | Each shim generated its own key; the spawn-race loser 401'd all session |
| `2e6af05f` | T14 | The shim that spawned the backend killed it on exit, cutting off sibling clients |
| `f8d4e0ba` | T16 | Hooks defaulted to a dev key the server never accepts; capture died silently |
| `28e69683` | T17 | Unquoted hook path broke on spaced home directories; matcher had the wrong shape |
| `f10225e5` | — | Install-time hook removal would delete a developer's hand-written hooks |
| `48feef30` | — | Env-supplied keys were never shared with hooks; a warming backend was misdiagnosed as a bad key |

The key-sharing mechanism is chosen once and used by both T15 and T16: a shared file at `<data-root>/.api-key`, written atomically by the shim, read by the hooks. Backend lifecycle is refcounted through pidfiles — last live shim out reaps, and a manually started server is never touched.

## Verification

Two of the three links in `shim persists → hook reads → server accepts` were executed against live processes, not stubs. A real shim on a fresh root generated and persisted a key; a real `memory-hook.ts SessionStart` process with the same root presented that key and completed all six briefing calls. Without the shared root it fell back to the dev key, got 401, and printed the loud banner — the failure is loud, which was the point.

The third link is source-verified rather than executed: `serverEnv["SHODH_DEV_API_KEY"] = API_KEY` feeds `configured_api_keys()` in `src/auth.rs`, which accepts that variable outside production. No server binary was built.

173 vitest tests pass (45 added); `bun run build` green.

**This machine cannot distinguish fixed from broken**, for three independent reasons found during verification: `.mcp.json` pins the dev key, the repo's `.env` sets the same key and Bun auto-loads `.env` from cwd, and the hook's legacy fallback is that same literal. Any future test of this path must run from a directory with no `.env`.

## Known limitations, deliberately not fixed

- If `SHODH_MEMORY_PATH` is set in an MCP `env` block but not in the shell, shim and hooks resolve different roots and the hook falls back to the dev key. Loud 401 rather than silence, but not working.
- `shodhDataRoot()` omits the Rust `default_storage_path()` legacy `./shodh_memory_data` branch. Harmless — the server never reads the key file — and the comment now says so instead of claiming parity.
- A recycled pid in `.mcp-shims/` reads as a live sibling and can pin a leaked backend. Costs a stray process, not data.

Full reasoning, evidence and unproved claims: `docs/agent-runs/2026-08-07-onboarding-first-five-minutes.md`.

### DIFF
--- a/docs/agent-runs/2026-08-07-onboarding-first-five-minutes.md
+++ b/docs/agent-runs/2026-08-07-onboarding-first-five-minutes.md
@@ -1,0 +1,107 @@
+# Onboarding: the first five minutes (T14–T17)
+
+Branch: `fix/onboarding-first-five-minutes` · Date: 2026-08-07
+
+Fixes the four defects that make a fresh install fail — three of them silently.
+Work was done by a Fable agent (commits `0481a926`, `2e6af05f`, `f8d4e0ba`,
+`28e69683`) and then independently verified, which turned up three further
+defects fixed in `f10225e5` and `48feef30`.
+
+## Why this was the priority
+
+A locally-deployable memory that nobody successfully installs is
+indistinguishable from no product. Of the four defects, T15 and T16 share one
+root — the auto-generated API key was never persisted, so nothing else could
+learn it — and both end in capture 401ing forever while the user believes
+memory is working. That is worse than a crash: a crash gets reported.
+
+## The mechanism, chosen once
+
+A shared key file at `<data-root>/.api-key`, written by the MCP shim and read by
+the hooks. Zero IPC, works for `bash` hooks and the standalone `memory-hook.ts`
+alike, and collapses the concurrent-spawn race because both shims converge on
+the same key. Creation is atomic (temp file + hard link, `wx` fallback), so a
+concurrent reader can never observe a half-written key.
+
+`<data-root>` is `SHODH_MEMORY_PATH`, else the platform data dir. It is **not** a
+faithful mirror of the Rust `default_storage_path()` — see below.
+
+## What each commit does
+
+| Commit | Defect | Failure mode it removes |
+|---|---|---|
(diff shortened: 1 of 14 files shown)


varun29ankuS:
Follow-up commit `265ae256`: `publishSharedApiKey` is now gated on the key's provenance as well as the backend URL. `isLocalServer()` alone is satisfied by a loopback tunnel or local proxy in front of a remote backend, so `SHODH_API_KEYS` — the production-shaped variable — was reachable for on-disk publishing. Restricted to `SHODH_API_KEY`/`SHODH_DEV_API_KEY`.

Also verified: `probeApiKey()`'s status classification works for the IPC transport, not just HTTP — `ShodhIpcClient` throws `IpcApiError` with the same `API error <status>` message shape, so a rejected key under IPC is reported rather than swallowed as inconclusive.

**CI has not triggered on this PR.** No workflow runs at all since 2026-08-06 10:33 UTC, across two pushes and PR creation. `ci.yml` on `main` has the correct `pull_request: branches: [main]` trigger, Actions is enabled with `allowed_actions: all`, and the workflow is active. Cause unknown — worth a look at the Actions tab for a banner before merging, since this PR should not merge unverified by CI.

PR #492, opened 2026-08-10
Break clear-text logging/storage taint paths; stop CSP clobbering the retired HTML surface

Portll:
CodeQL's security suite reports six results on `main` for the JavaScript/TypeScript surface — four `js/clear-text-logging`, one `js/clear-text-storage-of-sensitive-data` — plus one CSP defect on the retired HTML routes that isn't scanner-visible. This PR breaks each taint path at its narrowest point. A differential scan (same CLI, same queries, database rebuilt from each tree) goes **6 findings on `main` → 0 on this branch**.

## The flows and the fixes

**1. `mcp-server/index.ts` — stream handshake logged `USER_ID`** (`js/clear-text-logging`)
`process.env.SHODH_USER_ID` → `USER_ID` → `console.error` on WebSocket handshake. Operators put emails/names in `SHODH_USER_ID`, and MCP stderr ends up persisted in client logs. The handshake log now carries no id.

**2. `mcp-server/index.ts` — retry logger reprinted the id via error messages** (`js/clear-text-logging`)
`USER_ID` → endpoint templates (`/api/users/${USER_ID}/stats` and friends) → `backendRequest`'s invalid-JSON error embedded the endpoint → caught by the retry loop → `lastError.message` logged on every retry. Fixed where taint enters the message: the invalid-JSON error reports the HTTP status instead of the endpoint. All other errors reaching that logger were audited — none carries env-derived strings.

**3. `mcp-server/index.ts` — startup banner logged `USER_ID`** (`js/clear-text-logging`)
The banner now prints `claude-code` when the default is in use, or a "(custom, set via SHODH_USER_ID)" marker otherwise — still tells you which mode you're in, without persisting the identifier.

**4. `hooks/memory-hook.ts` — auth-failure report flagged transitively** (`js/clear-text-logging`)
`keyOrigin` never contains key material (it's a literal or a config-dir path), but it was derived by reading `.source`/`.file` off the same object that holds the resolved API key, so taint tracking flags it. Rather than suppress, the origin description is now its own module-level string assigned from literals at resolve time — clearer code, and the analyzer can prove it clean.

**5. `demo/spatial-map/index.html` — API key persisted to localStorage** (`js/clear-text-storage-of-sensitive-data`)
The key now lives in memory only; the page purges any key a previous version stored (`localStorage.removeItem` on load) and keeps persisting only the host. Cost: the demo asks for the key again per page load, with the host prefilled.

**6. `src/middleware.rs` + `src/handlers/visualization.rs` — CSP clobbers the retired pages' styling**
`security_headers` unconditionally overwrites `Content-Security-Policy` with `default-src 'none'`, which strips the inline `<style>` of the `/dashboard` / `/graph/view` moved-notice page — it renders unstyled. The middleware now inserts its CSP only when the handler didn't declare one, and the retired routes declare `default-src 'none'; style-src 'unsafe-inline'; base-uri 'none'; form-action 'none'`. API routes are byte-for-byte unchanged. A new test (`security_headers_keep_handler_declared_csp`) pins the non-clobbering behaviour.

## Verification

- Differential CodeQL (CLI 2.4.2 query pack, `CleartextLogging` / `CleartextStorage` / `FunctionalityFromUntrustedSource`): 6 → 0.
- `mcp-server` vitest: 200 passed on this branch; the 10 failures in `tests/ipc-client.test.ts` are identical on pristine `main` in the same environment (`listen EINVAL` binding Unix sockets under a sandboxed temp dir) — pre-existing/environmental, not from this change.
- `cargo test --lib middleware::`: 6/6 including the new regression test.
- `hooks/memory-hook.ts` parses clean (esbuild); its bun test suite covers formatting helpers this PR doesn't touch, and module-init order is unchanged (the description is initialized before `resolveApiKey()` runs).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,6 +149,43 @@ jobs:
           echo "" >> $GITHUB_STEP_SUMMARY
           echo "Single self-contained artifact: $SIZE bytes" >> $GITHUB_STEP_SUMMARY
 
+  seat:
+    name: Seat Harness
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/setup-node@v7
+        with:
+          node-version: "22"
+          cache: npm
+          cache-dependency-path: seat/package-lock.json
+
+      # seat/ was not compiled or tested by any workflow, so its type surface
+      # could break and every PR would still be green — the same gap the
+      # dashboard job above was added to close. It matters more than usual
+      # here: seat/src/backend.ts classifies backend failures into a closed
+      # vocabulary that GET /healthz publishes UNAUTHENTICATED, and the
+      # guarantee that only those literals can reach the response is enforced
+      # by the type checker. Nothing else enforces it.
+      - name: Install
+        working-directory: seat
+        run: npm ci
+
+      - name: Typecheck
+        working-directory: seat
+        run: npm run typecheck
+
+      - name: Build
+        working-directory: seat
+        run: npm run build
+
+      # Includes live-socket probes against closed ports and an unresolvable
+      # host: the classifier decodes shapes Node produces, and only a real
+      # socket proves those are the shapes it actually gets.
+      - name: Test
+        working-directory: seat
+        run: npm test
+
   feature-gates:
     name: Optional Features Compile
     runs-on: ubuntu-latest
@@ -843,7 +880,7 @@ jobs:
     # skipped on pull requests (if: push to main), so requiring
     # `result == 'success'` here would fail every PR.
     needs:
-      [version-check, format, clippy, build, test-unit, test-integration, security, frontend, feature-gates, benchmark]
+      [version-check, format, clippy, build, test-unit, test-integration, security, frontend, seat, feature-gates, benchmark]
     if: always()
     steps:
       - name: Generate final summary
@@ -861,6 +898,7 @@ jobs:
           integration_icon="${{ needs.test-integration.result == 'success' && '✅' || '❌' }}"
           security_icon="${{ needs.security.result == 'success' && '✅' || '❌' }}"
           frontend_icon="${{ needs.frontend.result == 'success' && '✅' || '❌' }}"
+          seat_icon="${{ needs.seat.result == 'success' && '✅' || '❌' }}"
           features_icon="${{ needs.feature-gates.result == 'success' && '✅' || '❌' }}"
           bench_icon="${{ needs.benchmark.result == 'success' && '✅' || '⚠️' }}"
 
@@ -872,6 +910,7 @@ jobs:
           echo "| Integration Tests (Tier 2) | $integration_icon ${{ needs.test-integration.result }} |" >> $GITHUB_STEP_SUMMARY
           echo "| Security Audit | $security_icon ${{ needs.security.result }} |" >> $GITHUB_STEP_SUMMARY
           echo "| Dashboard Build | $frontend_icon ${{ needs.frontend.result }} |" >> $GITHUB_STEP_SUMMARY
+          echo "| Seat Harness | $seat_icon ${{ needs.seat.result }} |" >> $GITHUB_STEP_SUMMARY
           echo "| Optional Features Compile | $features_icon ${{ needs.feature-gates.result }} |" >> $GITHUB_STEP_SUMMARY
           echo "| Benchmarks | $bench_icon ${{ needs.benchmark.result }} |" >> $GITHUB_STEP_SUMMARY
           echo "" >> $GITHUB_STEP_SUMMARY
@@ -884,6 +923,7 @@ jobs:
                 "${{ needs.test-integration.result }}" == "success" && \
                 "${{ needs.security.result }}" == "success" && \
                 "${{ needs.frontend.result }}" == "success" && \
+                "${{ needs.seat.result }}" == "success" && \
                 "${{ needs.feature-gates.result }}" == "success" ]]; then
             echo "### All critical checks passed" >> $GITHUB_STEP_SUMMARY
             exit 0
--- a/demo/spatial-map/index.html
+++ b/demo/spatial-map/index.html
@@ -210,7 +210,10 @@ body { font-family: 'Segoe UI', system-ui, sans-serif; background: #0a0a0f; colo
 
 <script>
 // ── Config ──────────────────────────────────────────────────────────────────────
-let API_KEY = localStorage.getItem('shodh_api_key') || '';
+// The API key is deliberately kept in memory only — never in web storage, where it
+// would persist in clear text. Purge any key a previous version of this page stored.
+localStorage.removeItem('shodh_api_key');
+let API_KEY = '';
 let HOST = localStorage.getItem('shodh_host') || 'http://localhost:3030';
 
 const MISSIONS = {
@@ -344,17 +347,13 @@ const LINEAGE_EDGES = {
 const RUN_TOKEN = Date.now().toString(36);
 
 // ── Setup ───────────────────────────────────────────────────────────────────────
-if (API_KEY) {
-  document.getElementById('inpHost').value = HOST;
-  document.getElementById('inpKey').value = API_KEY;
-}
+document.getElementById('inpHost').value = HOST;
 
 function connect() {
   HOST = document.getElementById('inpHost').value.trim() || 'http://localhost:3030';
   API_KEY = document.getElementById('inpKey').value.trim();
   if (!API_KEY) return;
   localStorage.setItem('shodh_host', HOST);
-  localStorage.setItem('shodh_api_key', API_KEY);
   document.getElementById('setup').style.display = 'none';
   loadAll();
 }
--- a/hooks/memory-hook.test.ts
+++ b/hooks/memory-hook.test.ts
@@ -4,6 +4,8 @@ import {
   formatRelativeTime,
   formatMemoriesForContext,
   isErrorOutput,
+  describeKeyOrigin,
+  reportAuthFailure,
 } from "./memory-hook";
 
 describe("formatRelativeTime", () => {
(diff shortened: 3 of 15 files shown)


Portll:
Added a third commit (0d371a3) covering a finding the full CodeQL suite surfaces beyond the original four queries: **js/stack-trace-exposure** in `seat/src/server.ts`.

The route catch-all serialized `error.message` from any unhandled exception into the response body — ledger fs errors carry file paths, and a failed backend probe can quote the URL it failed against, which the health route's own comment notes is sometimes where the credential lives. Non-`HttpError` failures now return a generic 500 with the real error logged server-side; `HttpError` messages (which are authored for clients) are unchanged. The unauthenticated health probe's `detail` collapses to `"unreachable"`, with the probe error logged.

Verification: `tsc --noEmit` clean for `seat/`; full-suite differential now **7 findings on `main` → 0 on this branch** (the five rules touched: CleartextLogging, CleartextStorage, FunctionalityFromUntrustedSource, FunctionalityFromUntrustedDomain, StackTraceExposure).

varun29ankuS:
John — the taint analysis is right and the differential scan is the correct way to demonstrate it. Six to zero with the database rebuilt from each tree is the standard I'd want. Three of the five changes I'd take as-is; two I'd like to talk about.

**The CSP fix is the best part of this PR.** The middleware was unconditionally overwriting handler-declared policy, so a browser-facing route could never set its own — that's a class of bug, not an instance, and making it a fallback with a test pinning the behaviour is the right shape.

**The spatial-map change is right, including the part you didn't have to do.** Purging a previously stored key rather than only stopping new writes is what makes it an actual fix rather than a fix for future users.

**`hooks/memory-hook.ts` I'd push back on.** Moving the origin description into a mutable module-level `KEY_ORIGIN_DESCRIPTION` trades a correct-by-construction derivation for one that depends on call ordering. Today it's derived from `RESOLVED_API_KEY.source` and cannot be stale; afterwards it's a global mutated as a side effect of `resolveApiKey()`, so if that's ever called twice — or `reportAuthFailure()` runs before it — the message is wrong. The tainted value is `.file`, not `.source`, so the narrower fix is to keep the derivation and drop the path from the message. Same CodeQL result, no order dependence.

**The seat health `detail` I'd like to keep richer.** We diagnosed a live problem yesterday off exactly that string — `Backend unreachable (GET /health): The operation was aborted due to timeout` told us immediately it was a timeout rather than connection-refused or DNS. Collapsing it to `"unreachable"` removes that signal. The route is authenticated and the field is operator-facing; could we classify rather than erase — `timeout` / `refused` / `dns` / `other` — so the taint breaks without losing the distinction?

**The MCP startup log is a smaller version of the same question.** `(custom, set via SHODH_USER_ID)` protects operators who put an email there, which is a real concern. But it also means a misconfigured id can't be spotted from the log. Would a short prefix or a hash keep it debuggable while breaking the flow?

Happy to merge once the hook and the health detail are settled — the rest is good.

Portll:
Hi Varun, I took both points onboard as they've both got some validity. One of them is actually stronger than you might have seen, and one rests on a premise the source contradicts. Detail below, plus a differential scan that caught a defect in my own follow-up.

Sorry I shipped a couple of more cosmetic fixes among the better ones.

Your framing was "same CodeQL result, no order dependence.", but after examinaion the version on this branch still logs the key file path:

```ts
// hooks/memory-hook.ts:72
KEY_ORIGIN_DESCRIPTION = `the shared key file ${file}`;
```

That string is printed at `:400` (stderr) and again at `:406` (the user-visible `systemMessage`). Moving the read off `RESOLVED_API_KEY.file` to a local `file` variable severed the dataflow edge without changing a byte of what reaches the log. CodeQL went 1 → 0 on provenance, not on disclosure. The path is `<data-dir>/.api-key` under `os.homedir()`, so it carries the operator's username — which is presumably why the query classified it.

Your fix is the only one of the two that actually closes it. Applied as you specified: derivation from `.source`, path dropped. Two changes on top:

- `switch` over the union with a `never` default, so a fourth `source` is a compile error rather than a silently missing branch, and an out-of-union value at runtime yields a sane string instead of `undefined`.
- A test pinning it — the same shape you asked for on the CSP fix. Nothing in the repo would have caught the path coming back; now `bun test` does, without waiting on a scan.

## The health route is not authenticated

`/healthz` is dispatched at `seat/src/server.ts:307`, before `this.authorize(request)` at `:312`. The comment at `:441` says "deliberately UNAUTHENTICATED" and `seat/README.md:137` documents it that way. `authorize()` is also a no-op when no token is configured (`:242-243`).

This doesn't change your conclusion, but it changes the reason, and the reason is load-bearing: if the route were authenticated, the obvious fix is to restore the full message behind the token. A refused connection reports `connect ECONNREFUSED 127.0.0.1:45999` — host and port. What makes a classification safe to publish is that it's a fixed set of literals, not the trust level of the caller.

Another version of your ordering argument is that `resolveApiKey()` never *resets* `KEY_ORIGIN_DESCRIPTION` to its default. It only escalates. There's no ordering of two calls that can lower it again.

## The classification can't live in the handler

I tried it there first. It can't work, because it also corrects the incident you're remembering.

`ShodhBackend.request` flattens the error before the handler sees it (`backend.ts:262-265`), and `ShodhBackendError` carries only `status` and `body`. `.cause` — which is where the code lives — is discarded. I extracted three candidate classifiers and ran them against real sockets through a verbatim copy of that wrapper:

| case | read `.code` in server.ts | match message in server.ts | classify at throw site |
|---|---|---|---|
| timeout | other | timeout | timeout |
| refused | other | other | refused |
| dns | other | other | dns |
| tls | other | other | tls |
| **correct** | **0/4** | **1/4** | **4/4** |

For example:

```
refused -> "Backend unreachable (GET /health): fetch failed"
dns     -> "Backend unreachable (GET /health): fetch failed"
```

`backend.ts:263` takes `error.message`, and for every transport failure except a timeout that's just `fetch failed`. Timeouts survive because `DOMException`'s message describes itself, which is exactly why your incident was diagnosable. Had it been refused or DNS, the log would have read `fetch failed` for both, so the  distinction you want preserved never existed for two of the three classes you named. Classifying at the throw site doesn't restore it; it creates it.

So: `classifyTransportError` in `backend.ts`, walking `.cause` and `AggregateError` (dual-stack `localhost` nests one), and `kind` as a **required** constructor argument so no throw site can omit it. `HealthDetail` is typed `BackendFailureKind | \`http-${number}\`` — interpolating a hostname into that field is a compile error, not a code-review convention.

One branch you didn't flag and I'd removed: `handleHealth` caught HTTP errors too, so a backend answering 500 reported `unreachable`. It's reachable; it's erroring. That's now `http-500`.

## Differential scan

Rebuilt the database from each tree, CLI 2.26.1, query pack 2.4.2, same three queries. **6 → 0, with 0 new.**

Two corrections to the PR description, both mine:

- The breakdown says four `js/clear-text-logging`. It's **five**, plus the one storage finding — six scanner-visible. The CSP defect is *additional* to those, not one of the six. `hooks/memory-hook.ts` accounts for two (`:395` and `:400`), which independently confirms the key origin reached two sinks.
- The bun hook suite is 6/13 red on `main` — stale tests against a changed `formatMemoriesForContext` signature, unrelated to this branch. Repaired here rather than left; it's 23/23 now, including the `describeKeyOrigin` guard and a test on what `reportAuthFailure` actually emits on both streams. One behavioural fix went in with it: `isErrorOutput` missed lowercase `error:`, which is what cargo emits. `FAILED` stays case-sensitive deliberately — relaxing it makes `0 tests failed` a false positive, so I've pinned both trades with tests.

**The scan caught a defect in my own follow-up.** I'd extracted the startup-banner logic into `describeUserId()` and included the id's length for misconfiguration detection, on the reasoning that the value is hashed. Hashing sanitizes; `.length` does not, and it carried `process.env` through to `console.error` — a new `js/clear-text-logging` at `index.ts:6926`. Your inline ternary was clean so my improvement broke it. Removed — the digest already changes on a trailing newline, so it added nothing. It also moved to `index-helpers.ts`, since that is where this module keeps anything unit-testable; in `index.ts` it had no coverage at all, which is how it got through.

To your prefix-vs-hash question: hash. A prefix leaks the local-part, which is the case you're protecting. `custom (sha256:ff8d9819)`, reproducible with `echo -n "$SHODH_USER_ID" | shasum -a 256 | cut -c1-8`. Also to be clear, against a low-entropy input this is obfuscation, not anonymisation — it defeats incidental disclosure in a shared log, which is the threat you're probably most concerned with (otherwise move away from SHA256).

## Verification

Differential CodeQL, database rebuilt from each tree: **6 → 0, nothing new**. CLI 2.26.1, query pack 2.4.2, same three queries.

- `cargo test --lib errors::` — 5/5
- `seat` typecheck + build clean; **24/24** tests — synthetic error shapes, live sockets (closed port, unresolvable host, blackhole IP), and an end-to-end suite that runs a real `SeatServer` against a real backend made to fail six ways
- hooks — **23/23** under `bun test`
- `mcp-server` helpers — **20/20** under vitest, 6 of them new on `describeUserId`
- 3 compile-time guards confirmed firing (`TS2322` on a hostname in the published field, `TS2673` on external construction, `TS2345` on an invented kind)
- 5 mutation tests confirming the suite fails when the implementation regresses

72 tests total, against 13 (6 of them failing) on `main`.

## Four loss scenarios, closed in code

Ran an STPA pass over the result rather than shipping the caveats. Four unsafe control actions were still reachable; all four are fixed here.

**The `USER_ID` round-trip, closed at the source.** `AppError::client_message` gated 5xx sanitisation on `is_production_mode()`, which is `unwrap_or(false)` — so outside production the full internal message went into the response body, and `src/handlers/state.rs:1302` formats the user id into exactly such a message. The MCP shim embeds response bodies in its retry log (`index.ts:690` → `:935`), so the id round-tripped into client logs. No static analysis can see this: the value leaves the process and comes back over the wire.

5xx bodies are now sanitised unconditionally. Nothing is lost — `IntoResponse` already logs the full detail server-side at `error!` *before* sanitising, which is where a developer should be reading it. 4xx still returns verbatim. Two tests in `errors.rs`; `cargo test --lib errors::` is 5/5.

**A classification can no longer contradict its cause.** A required `kind` parameter forces a throw site to make a choice, not a correct one — `"other"` compiles fine. The constructor is now private, with three factories: `.transport(message, cause)` derives the kind by classifying, `.http` and `.protocol` fix their own. External construction is `TS2673`.

**The health route no longer names 4xx.** My own addition: `http-401` would have told an anonymous caller that the seat's backend credential is being rejected — a fact about our configuration, not the backend's liveness, on a route that answers before `authorize()`. 4xx collapses to `http-client-error`; 5xx keeps its status, since "up and broken" is the distinction operators need and a server-error code discloses nothing.

**DNS classification no longer assumes one resolver.** `EAI_NONAME`, `EAI_FAIL`, `EAI_NODATA` and `ENODATA` join `ENOTFOUND`/`EAI_AGAIN`. A sandboxed CI resolver reporting a different errno was going to fall through to `other`, in a test written specifically to catch that class of assumption.

## Scope options and open decisions

**The CI job.** `seat/` was compiled and tested by nothing — and `CI Summary` is the only required check, so a job not in its `needs` can't block a merge. I added `seat` to all three places. IMO It's the right fix and it's also infrastructure scope on a logging PR. `seat` typechecks clean on your head commit, so it won't drag in pre-existing debt, but if you'd rather this landed separately it's the last commit and lifts out cleanly. Without it, the compile-time guarantees above are local-only.

**The Rust commit.** Sanitising 5xx bodies unconditionally is the one change here that fixes a defect neither my scanner nor your review found, and it touches behaviour outside the JS surface. If you'd rather it went in on its own so it can be reviewed against the Rust side properly, it's a self-contained commit. I'd argue for keeping it: it crosses a network boundary into client logs by default, which makes it a worse leak than three of the six original findings, which were all local stderr.

**One thing I can't verify from here:** the CI job has never executed on GitHub. It's correct structurally — job defined, listed in `summary.needs`, present in the exit gate — and `seat` typechecks clean on your head commit so it won't drag in unrelated debt. But its first real run will be on this PR.

PR #499, opened 2026-08-13
docs: worktree rule in CLAUDE.md + hardware-path research

varun29ankuS:
**Two MUST rules in `CLAUDE.md`** requiring worktrees to live in `<repo>/.claude/worktrees/`, with the reasons stated rather than asserted:

- A worktree outside `kalki-v2/` never sees the parent `.cargo/config.toml`, which sets `build.target`. It therefore writes to `target/debug` instead of `target/x86_64-pc-windows-msvc/debug`, **forking `target/` into two complete trees** — that fork was 14.69 GB of a 46 GB `target/`.
- Branch refs survive `git worktree remove`, so **untracked files are the only thing it can destroy**. The removal rule says to check for them first. One worktree here held 291 untracked lines that existed in no branch and no git object.

The rule previously existed only in conversation, which meant it died at every session boundary and was never present in a dispatched subagent's context. Ten worktrees had accumulated across four roots.

**Plus** the hardware-path research document — an assessment of whether agent memory should become silicon, with every external number labelled MEASURED / SIMULATED / CLAIMED and internal claims verified against source at the cited `file:line`.


varun29ankuS:
Closing — splitting this. The CLAUDE.md worktree rule will come back as its own PR; the hardware-path research document is not going public.

PR #506, opened 2026-08-14
build: stop emitting dependency debuginfo in dev and test profiles

varun29ankuS:
## What

Adds the `[profile.dev]` this crate never had, and stops shipping debug info for dependencies.

```toml
[profile.dev]
debug = "line-tables-only"

[profile.dev.package."*"]
debug = false
```

Same settings restated on `[profile.test]` so a later `[profile.dev]` edit cannot silently change what `cargo test` emits, plus a `dev-debug` profile to opt back into full debuginfo when a real debugger session is actually needed.

## Why

Cargo's default is `debug = true` for the workspace *and* every dependency. With no `[profile.dev]` present, that default was in force. Measured on this machine today:

| | size |
|---|---:|
| `target/x86_64-pc-windows-msvc/debug` | 29.7 GB |
| — `deps/*.pdb` (7 files) | 3.46 GB |
| — `deps/*.rlib` (715 files) | 5.97 GB |
| — `build/` (native build-script output) | 9.78 GB |

Most of that rlib weight is dependency debug info that nothing in this repo reads. Nobody steps through rocksdb.

## Cost

Line tables are retained for workspace code, so backtraces and test failures still resolve to file and line. What is given up is variable inspection under a debugger — recoverable with `cargo build --profile dev-debug`.

Changing profile settings invalidates the build cache, so the first build after this merges is cold.

## Verification

`cargo metadata --no-deps` parses. A full clean rebuild on the new profile is running locally and the reclaimed size will be posted here.


varun29ankuS:
Cold rebuild on the new profile: exit 0, all test binaries link.

| | before | after |
|---|---:|---:|
| `debug/` total | 29.72 GB | 10.28 GB |
| `deps/*.rlib` | 715 files, 5.97 GB | 360 files, 1.29 GB |
| `build/` | 9.78 GB | 0.75 GB |

The headline drop overstates the profile's contribution: the 29.72 GB tree had accumulated across days and branches, while this is a single clean build, so some of it is de-accumulation that any `cargo clean` would have given.

The clean attribution is per-artifact — **average rlib 8.35 MB -> 3.58 MB, a 57% reduction**. That is the profile change and nothing else.

PDBs are now the largest single category in `deps` at 3.24 GB across 46 files. Those are workspace test-binary line tables, which is what we chose to keep.

PR #518, opened 2026-08-28
security: confine per-user paths to their base directory, and stop printing the full API key

Portll:
## Summary

Every per-user directory in the codebase is `base.join(user_id)` where `user_id` arrives from HTTP/MCP/CLI callers. A value like `../other-user` (or an absolute path) escapes the base directory at every one of those joins — path traversal into, and out of, other tenants' directories, in both the state manager and the backup engine (backup **restore** included).

Your `validation::validate_user_id` already states exactly the right contract — but it is called in some HTTP handlers and skipped in others, while `handlers/state.rs` joins `user_id` into filesystem paths at 15 places with **no validation call anywhere in the file**. So this PR does not introduce a second validator: `src/path_guard.rs::sanitize_component` **delegates to `validate_user_id`** and becomes the chokepoint two guarded helpers route every user-scoped join through — `ShodhBackupEngine::user_backup_dir` and the state manager's `user_dir`. An unvalidated id can no longer reach a filesystem path regardless of which handler it arrived by, and the two layers cannot drift because there is only one rule.

**Compatibility:** because the guard is your own validator, every id the API layer accepts today (including Unicode-alphanumeric ids) keeps working at the filesystem layer — no existing on-disk user directory becomes unreachable. It rejects rather than normalizes, since silently rewriting `../x` would hide the attempt.

Separately, `shodh init` printed the freshly generated API key to stderr, which lands it in every captured transcript and CI log. It now prints the first 12 characters and points at the config file that already stores it with restricted permissions.

## Origin

Found by CodeQL's Rust analysis (`codeql/rust-queries`: `rust/path-injection` ×35, `rust/cleartext-logging`) running as part of a scheduled scanner fleet over a downstream deployment of this codebase. The same fix is applied and verified there.

## Verification

- `cargo check` clean on this branch.
- `path_guard` tests assert **parity with the API layer** (Unicode ids accepted; traversal, separators, NUL, dot-names, over-length rejected) so the two boundaries cannot disagree.
- Existing lib tests compile and pass.

### DIFF
--- a/src/backup.rs
+++ b/src/backup.rs
@@ -84,9 +84,10 @@ impl ShodhBackupEngine {
     /// path in it would escape `backup_path` at every join below. Every user-scoped path in this
     /// engine goes through here.
     fn user_backup_dir(&self, user_id: &str) -> Result<PathBuf> {
-        Ok(self
-            .backup_path
-            .join(crate::path_guard::sanitize_component(user_id, "backup user id")?))
+        Ok(self.backup_path.join(crate::path_guard::sanitize_component(
+            user_id,
+            "backup user id",
+        )?))
     }
 
     /// Create a full backup of a RocksDB database
--- a/src/backup.rs
+++ b/src/backup.rs
@@ -79,6 +79,16 @@ impl ShodhBackupEngine {
         &self.backup_path
     }
 
+    /// The per-user backup directory, with the identifier validated as a single path component
+    /// BEFORE it is joined — `user_id` arrives from outside the process, and `../` or an absolute
+    /// path in it would escape `backup_path` at every join below. Every user-scoped path in this
+    /// engine goes through here.
+    fn user_backup_dir(&self, user_id: &str) -> Result<PathBuf> {
+        Ok(self
+            .backup_path
+            .join(crate::path_guard::sanitize_component(user_id, "backup user id")?))
+    }
+
     /// Create a full backup of a RocksDB database
     ///
     /// # Arguments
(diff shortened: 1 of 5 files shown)


varun29ankuS:
Thanks for this — the path confinement and the API-key redaction are both worth having, and the h2 bump was overdue on our side.

Two things before I merge:

1. `main` now carries h2 0.4.18 (via #514), so this will need a rebase — your 0.4.16 conflicts with it. Same fix, newer patch.

2. The lockfile also downgrades `socket2` 0.6.1 → 0.5.10. `cargo audit` is clean on it, so nothing is flagged, but could you say what that's for? We were bitten once by a dependency *downgrade* riding along inside an otherwise-good PR — a group update walked `lz4_flex` back into RUSTSEC-2026-0041 (CVSS 8.2) and it very nearly merged on green. Since then we ask about every downgrade, even the benign ones. If it's incidental to the h2 resolution, dropping it from the lockfile would let this go in on the security change alone.

Portll:
Rebased onto main (`8eb277e`) — both points addressed:

**h2:** the branch no longer touches it. Main's 0.4.18 (#514) clears RUSTSEC-2026-0258 upstream of this PR, so the audit-gate commit reduces to its rustfmt half.

**socket2:** your instinct was right and the downgrade was incidental. Came via produced the h2 0.4.16 pin; nothing in this PR uses socket2. The rebased branch has a **zero-line Cargo.lock diff against main** (`git diff main -- Cargo.lock` is empty), which is a stronger property than explaining the downgrade: there is now no dependency movement to reason about.

The security change stands alone, as you suggested.

Verified after rebase: `cargo check` clean, `path_guard` parity tests 2/2.

PR #519, opened 2026-08-28
deps: cargo update — h2 and every yanked pin clear; lz4_flex needs its own manifest bump

Portll:
Lockfile-only advisory sweep, measured by cargo-audit before/after:

- **h2 0.4.12 → 0.4.19** — clears RUSTSEC-2026-0258, the one *error* your Security Audit gate fails on (also fixed independently inside #518's lockfile; whichever merges first, the other rebases clean).
- Every **yanked**-crate warning clears (chacha20, spin ×2, stabby) — the update re-resolves them to non-yanked releases.
- Survivor: **RUSTSEC-2026-0041 lz4_flex@0.10.0** (patched ≥0.11.6) — past the manifest's 0.10 constraint, needs an API-reviewed bump rather than a lockfile move. The unmaintained/unsound advisories (bincode, paste, atomic-polyfill, event-listener, lru, rand) similarly have no in-constraint fix and stay visible as allowed warnings.
- `ort` deliberately held at 2.0.0-rc.11: rc.13 is source-incompatible (Error-type change breaks `.context()` call sites) — verified by compiling both.

Verification: `cargo check` clean, `cargo fmt --check` clean, `cargo audit` output as described.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -59,9 +59,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.4"
+version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
+checksum = "c982642fa9e8606056828ee9a8505737230110bb1099153c79efe865c59d12ba"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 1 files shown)


varun29ankuS:
Superseded by #539. Both of these carry `lz4_flex` 0.11.6 → 0.10.0 and `twox-hash` 2.1.3 → 1.6.3 — that is RUSTSEC-2026-0041 (CVSS 8.2) coming back. Neither PR is at fault: a plain `cargo update` on main does the same thing, because `zenoh-transport` 1.10.0 requires the older `lz4_flex` while `tantivy` correctly uses 0.13.1. #539 takes 212 packages forward and pins zenoh at 1.8.0 so the downgrade cannot occur.

PR #522, opened 2026-08-29
A truncation through an equal-score plateau is not a ranking, it is a coin flip

varun29ankuS:
## What went red

Three unrelated PRs — #506 (`+30/-0`, dev-profile debuginfo), #498 (SIMD dispatch) and #497 (stage timing) — all failed `L1 Smoke Suite` simultaneously with **`EXIT: 2`**, the harness's *infrastructure failure* code. Not a recall regression. A build-profile change with no logic in it cannot regress recall, which is what pointed away from the PRs.

From the `recall-eval-report` artifact:

```
kind: infrastructure
non-determinism [mode=full]: case conv-42_q19 rank list diverged
  between repeat 0 and repeat 1
  repeat 0 = [... "conv-42:D13:18", "conv-42:D8:6", "conv-42:D2:14", "conv-42:D13:19"]
  repeat 1 = [... "conv-42:D13:18", "conv-42:D2:14", "conv-42:D8:6", "conv-42:D13:19"]
```

Ranks 8 and 9, same two items, swapped. Repeats 0 and 3 agree with each other; 1, 2 and 4 agree with each other. It is a coin flip, not a drift.

## Why

Four sorts order by score alone and then truncate. Rust seeds every `HashMap` independently — two maps in the *same process* iterate in different orders — so when a truncation lands inside a plateau of equal scores, which element survives is decided by iteration order.

The harness already guards this class: RH-11 pins rank lists, RH-12 pins leg membership. And the fix was already applied to the memory sort inside the graph leg, with a comment naming this exact mechanism:

> `activated_memories` HashMap iteration order — random per process AND per ingest — and the truncations below cut through equal-score plateaus, so leg MEMBERSHIP would differ between two ingests of the same corpus (RH-12 guard).

That reasoning applies verbatim to the **entity reachability sort one layer upstream**, which collects straight out of a `HashMap` and truncates to `REACH_MAX_ENTITIES`. It was missed there, and at three other cut points.

## The change

| site | order now |
|---|---|
| reachability injection | score desc → entity id asc |
| multi-hop beam cut | score desc → endpoint id asc |
| BM25 bridge expansion | strength desc → bridge name asc |
| facts by entity | confidence desc → fact id asc |

**This cannot change ranking quality.** It only orders elements that compare *equal*, which previously had no defined order at all. No threshold moved, no assertion weakened, no tolerance widened — the gate stays exactly where it is, and the harness's own determinism check is the verification.

## Verification

- `cargo check --all-targets` clean
- `cargo clippy --all-targets` clean
- `cargo fmt --all` clean
- The real test is `L1 Smoke Suite` on this PR: it exercises `mode=full` at `REPEATS: 5` with the RH-11/RH-12 guards, which is precisely what caught the defect.

## Note on scope

I could not reproduce locally — the gate needs the LoCoMo corpus, the MiniLM and GLiNER weights, and five repeats. So this is a principled fix for the identified class rather than a bisected fix for one call site. If CI stays red, the surviving non-determinism is somewhere these four sorts don't reach, and the artifact will say where.

### DIFF
--- a/src/graph_memory.rs
+++ b/src/graph_memory.rs
@@ -4547,7 +4547,17 @@ impl GraphMemory {
         // Sort by pruning priority: LTP-protected edges last (survive pruning),
         // then by effective strength descending (strongest survive).
         //
-        // CROSS-INGEST DETERMINISM — the same defect Phase 3.1 of
+        // CROSS-INGEST DETERMINISM. Latent, not observed: this was fixed while
+        // hunting the `conv-42_q19` repeat divergence and it did NOT fix it.
+        // Measured on the locomo gate before and after, the metrics are
+        // bit-identical (ndcg@10 0.4114 both runs) and the divergence survives,
+        // so no entity in that corpus exceeds MAX_ENTITY_DEGREE and this branch
+        // never executes there. It is kept because the defect is real on any
+        // corpus with hubs — which the defence and GDELT graphs have — not
+        // because it was shown to matter here. The remaining divergence is
+        // elsewhere and is present on main independently of this branch.
+        //
+        // It is the same defect Phase 3.1 of
         // `get_entity_relationships_limited` fixes on the READ path, which this
         // write path never got. `get_entity_relationships` returns edges in
         // prefix-scan order, i.e. edge-UUID order, and edge UUIDs are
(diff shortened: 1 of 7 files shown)


varun29ankuS:
Superseded by #538, which is the same change replayed onto current main. #522 reported 3,800 insertions / 3,712 deletions; the real change is 121 / 33 — the rest was `src/handlers/recall.rs` rewritten from CRLF to LF end to end. #538 shows only the real change and keeps the file's line endings intact.

PR #546, opened 2026-09-10
Count index-insert loss, the half the commit guard does not cover

Portll:
`BM25Index::commit_failures` exists because a lost commit batch once made 432/432 smoke comparisons diverge silently, and `run_smoke_suite` bails on it. This adds the half that guard does not cover.

## The gap

A lost **commit** means the batch never reached disk. A lost **insert** means the document never reached the writer — the memory is stored in the primary store and no query can reach it. Same class of damage, and only the first is counted.

Both are handled-and-continued at eleven sites in `memory/mod.rs`:

```rust
if let Err(e) = self.retriever.index_memory(&memory) { /* log, continue */ }
if let Err(e) = self.hybrid_search.index_memory(...) { /* log, continue */ }
```

That is correct for a server, which should keep serving. It is wrong for an eval, which is then about to report recall over a corpus smaller than the one it ingested — and today it does so without noticing.

The vector side has the same hole: a failed Vamana insert leaves a memory stored and semantically unreachable. That is the orphan condition the startup recovery already repairs, which means the condition is known — it has just only ever been observable *after* a restart notices, never while it accumulates.

## What this does

- `index_failures` beside `commit_failures` on `BM25Index`, and the equivalent on `RetrievalEngine`.
- Incremented **at the boundary** (inside `upsert` / `index_memory`), not at the eleven call sites, so a caller added later cannot forget to record its own loss.
- `MemorySystem::index_failure_counts() -> (vector, lexical)`.
- `run_smoke_suite` bails on a nonzero count exactly as it already bails on lost commits. This is the point of the change: the counter is load-bearing rather than decorative.

Nothing changes in the server's behaviour — the sites still log and continue, they are now merely countable.

## Verification

The test drives a real failure rather than asserting the wiring: it removes the index directory under the writer, then asserts that the error and the count **agree in both directions**. It fails if an insert errors uncounted, and equally if the count moves while the insert succeeded. A counter nobody has watched move is the defect it exists to catch.

Measured with a control arm rather than a bare pass:

| | result |
|---|---|
| pristine `a292546` | 1140 passed, 0 failed, 9 ignored |
| this branch | 1141 passed, 0 failed, 9 ignored |
| delta | exactly the one test added |

Both arms were built into their own `CARGO_TARGET_DIR`. On a machine with a shared target dir, two worktrees produce the same binary path (`package+version+features`), so one build overwrites the other and a suite can silently execute another tree's code — a green run stops being evidence about the tree you are in. Flagging it because it cost a colleague a debugging cycle here this week, and it is invisible when it happens.

## Provenance

Found while porting this harness into a downstream fork, where the eleven-site pattern is identical. The fork carries the same fix.


varun29ankuS:
Verified the mechanism and it's a real gap worth closing. `commit_failures` covers a batch that never reached disk; nothing covered a document that never reached the writer. Same damage — stored and unreachable — and only one was visible.

Two things I checked rather than took on trust, both good:

**The boundary placement holds.** BM25 wraps the actual `add_document`; the vector side wraps `index_memory_inner`. A path added later can't forget to count, which is the property that matters and the reason the exact call-site count doesn't.

**The harness gate is load-bearing.** `run_smoke_suite` bails and says why it's an infrastructure failure rather than a retrieval result. That's what keeps the counter from being decorative.

I also reproduced your control arm on this machine: 1141 passed / 0 failed / 9 ignored on the branch, matching what you reported.

## One change before merge: the test passes without exercising the counter

```rust
assert_eq!(result.is_err(), index.index_failure_count() == 1, ...)
```

That is satisfied when both sides are false. I instrumented it:

```
PROBE: err=false count=0
```

`remove_dir_all` doesn't induce a failure — tantivy holds the directory open — so the insert succeeds, the counter never moves, and the test proves nothing. Your own line is the standard it misses: *"a counter nobody has watched move is the defect it exists to catch."*

The fix is to drive a real failure — drop or close the writer first, or feed a schema-invalid document — and assert `result.is_err() && count == 1` directly. Agreement-in-both-directions is a good instinct, but it needs the failure to be guaranteed first.

Minor: "eleven sites" is thirteen in `memory/mod.rs`. Harmless, and in the safe direction.

## Separately

The shared-`target/` warning is right and worth more than a footnote — it cost us a debugging cycle here recently. Two worktrees produce the same binary path, one build overwrites the other, and a suite silently executes another tree's code; a green run stops being evidence about the tree you're in. Good catch, and the control-arm methodology is the right way to report a test-count delta.

Portll:
Thank you — the probe is the finding. `assert_eq!(result.is_err(), count == 1)` passes on `false == false`, so the test asserted an agreement that was never exercised and the counter was never watched move. It has been deleted rather than repaired, for a reason worth stating, and replaced with something stronger.

**The suggested induction is not reachable through tantivy's public API.** In 0.25.0, `IndexWriter::add_document` hands the document to `send_add_documents_batch`, which returns an error on exactly one condition: `index_writer_status.is_alive()` is false — "An index writer was killed". Nothing else on that path fails. Dropping the writer moves it rather than killing it, closing it consumes `self` (so the handle is gone and there is nothing left to call `add_document` on), and `SegmentUpdater::kill` is private and takes `&mut self`. A schema-invalid document does not help either: the document is built against the schema, and the failure that does exist surfaces at commit, which is what `commit_failures` already counts. The insert counter is therefore narrow by construction on the BM25 side — it can only fire when the writer is already dead — and I have documented that on the field instead of leaving a reader to assume the counter is broadly load-bearing. The vector-side counter is not narrow in the same way; embedding and vector-insert failures are ordinary.

**The replacement: reconcile populations rather than trust that every loss announced itself.** `MemorySystem::index_coverage()` returns `(stored, vector_indexed, lexically_indexed)` from `get_all_ids().len()`, `RetrievalEngine::len()` and `bm25_doc_count()` — all three already existed. That is strictly stronger than the counters, because it does not depend on a loss producing an `Err` for anything to observe. It catches an insert that returned `Ok` having done nothing, a path that never indexed at all, and a commit that vanished silently — none of which increment a counter. `ingest_corpus` now bails on a shortfall after the two existing guards, shortfall-only so that a legitimate surplus (chunking, re-index) cannot false-fail.

**The new test is verified failing, which is the part the deleted one lacked.** `every_stored_memory_is_present_in_both_indexes` remembers 8 experiences and asserts `stored == vector_indexed == lexically_indexed == 8`; I measured that equality on this path before asserting it rather than assuming it. To show it is load-bearing I mutated a silent drop into `RetrievalEngine::index_memory` — return `Ok(())` without inserting, no error raised — and the test fails with `vector index holds 3 of 8 stored memories`, while both failure counters read 0. That mutation is precisely the case the counters cannot see, and it is the case the reconciliation exists for. Reverted after measuring.

**On the count: your thirteen is right and my eleven was the same defect as the test** — I grepped one syntactic form and reported the population. The full decomposition, now stated in the code: thirteen `index_memory` sites (eleven `if let Err`, two `match`) plus two more via `reindex_memory` — fifteen index-insert sites in `memory/mod.rs`, and I checked each body: none propagates, all fifteen log and continue.

**Control arm, restated honestly.** My earlier "1140/1141" was the lib target only, not the whole suite — the lib is 1140 of roughly 1500. Worse, one intermediate full-suite run of mine was killed by a `timeout` and I summed its partial output into a clean-looking total before noticing the binary count was 10 of ~40; the exit code carried the whole signal and a pipeline had eaten it. Re-run unbounded, the numbers below are the full suite. The net test delta is zero by construction: one lib test deleted, one integration test added.

Full suite at `6f97469`, 40 binaries, isolated `CARGO_TARGET_DIR`: **1969 passed, 2 failed, 12 ignored**. The two failures are `timing_sla_tests`' wall-clock thresholds — `test_sla_record_p50_latency` at 572ms against a 500ms bound and `test_sla_throughput_sustained` at 3.6 rec/s against a 5 rec/s floor — measured while the rest of the suite ran in parallel on a loaded machine. That binary is **13/13 green run alone** (`--test-threads=1`), and nothing in this change touches the `remember()` path: `index_coverage` is a read-only accessor whose only callers are the new test and the harness guard. Flagging them rather than filtering them out, since a suite I report as green should be one I actually saw go green. Per-target: lib 1140 (unchanged from base — the deleted test is the one the parent commit added), `memory_persistence_tests` 21 against base's 20.

One note on the p50 tests independent of this PR: under contention a median is the wrong statistic for a latency floor, because a handful of descheduled samples move it. A minimum, or a p50 taken over repeated runs rather than within one, would be stable against exactly the load that broke these two. Happy to open a separate issue if that is useful.

varun29ankuS:
The reconciliation is the right replacement. Comparing populations catches the loss that never returned an `Err` — the case the counters structurally can't see — and the mutation result is the evidence the first test lacked. I checked what it depends on: `RetrievalEngine::len()` counts memories, not chunks (`IdMapping::memory_to_vectors` keys); `remember()` commits and reloads BM25 per call, so the count isn't racy; and the dead-writer-only behaviour of `add_document` holds in 0.26.1, which is what we lock (you cited 0.25.0).

Four small things before merge. **CI Summary is the one required check and it depends on Format Check**, so (2) blocks the merge on its own:

1. **`retrieval.rs`**: `index_failure_count` sits between `index_memory`'s doc block and `pub fn index_memory`, so the chunking/atomic-storage docs render on the counter and `index_memory` has none. Move the fn above that doc block. It's also worth saying on the counter that it covers `index_memory` and `reindex_memory` (which delegates), while the startup migration and orphan-recovery paths call `add_vector` directly and aren't counted.
2. **Format Check**: `cargo fmt` wraps the `assert_eq!(stored, N, …)` line in `every_stored_memory_is_present_in_both_indexes`.
3. **`hybrid_search.rs` `upsert` body**: still says "eleven call sites"; the field doc now says fifteen.
4. **`runner.rs` coverage-guard comment**: it justifies shortfall-only with "a surplus (chunking, re-index) is not loss", but all three counts are per memory, so chunking can't produce a surplus. The real reason is that an index entry with no stored memory is a stale index, a different defect from the one this guards.

Once CI Summary is green I'll merge. L1 is red on main for an unrelated recall regression, so I won't count it against this PR.

On the SLA p50 tests: agreed, a median under contention measures the scheduler. A separate issue would be welcome.

Portll:
Rebased on current main (`a724b86`) and addressed all four:

1. `index_failure_count` now sits above `index_memory`'s doc block, so the chunking/atomic-storage docs render on `index_memory` again. Its own doc says it covers `index_memory` and `reindex_memory` (which delegates), and that the startup migration and orphan-recovery paths call `add_vector` directly and are not counted.
2. `cargo fmt --all` — the one diff was the `assert_eq!` you named. `cargo fmt --all -- --check` is clean on the branch.
3. `upsert` comment says fifteen, matching the field doc.
4. Coverage-guard comment now gives the per-memory reason: all three counts are per memory so chunking cannot produce a surplus, and an index entry with no stored memory behind it is a stale index, a different defect from the one this guards.

Re-run on the rebased branch: `every_stored_memory_is_present_in_both_indexes` passes, and clippy at CI's flags exits 0.

Portll:
One more thing on CI: Security Audit is red on every branch since this morning for rustls 0.23.43 (RUSTSEC-2026-0285, in the lock, not this PR). #554 is the three-line lockfile bump for it; this branch will rebase onto it once that lands.

PR #549, opened 2026-09-13
fix(storage): search_by_tags returns ids in a stable order, so a truncated tag recall is repeatable

Portll:
## What
`search_by_tags` collected matches into a `HashSet` and returned its iteration order; `recall_by_tags` truncated that to the caller's limit. With more matches than the limit, *which* records came back changed between identical calls. The storage layer now returns ids sorted, and `recall_by_tags` sorts the materialised memories newest-first before it truncates, so the page `/api/recall/tags` serialises is the newest `limit`, not a stable-but-arbitrary `limit`. `search()`'s `Combined` arm sorts after intersecting, since its `HashSet` discarded whatever order the leaves produced.

## Why
Measured on a fork with ≥5000 records under one project tag against the MCP default limit of 50: two identical `/api/recall/tags` calls returned different sets, not a different order over one set. `sessions.rs` already worked around this by fetching everything and sorting; the endpoint has no such caller-side fix, and its client gets storage order verbatim.

## Not in this PR
No change to the AND/OR semantics of multi-tag search (it is a union). No blanket sort in `search()`: the by-date and by-location scans return a meaningful order that `recall_by_date` truncates on, and an id sort there would replace oldest-first with arbitrary.

## Verification
`recall_by_tags_page_is_repeatable_and_newest_first` seeds 40 under one tag, asks for 10 twice, and asserts the same set both times and that it is the newest 10 of 40. With both sorts reverted to main: it fails on the same-set assertion. The storage-level sort alone has run on the fork's production store since 2026-09-09. `cargo fmt --all -- --check` clean on the rebased branch.

Found by commitwork.


varun29ankuS:
The premise checks out end to end: `search_by_tags` returns `HashSet` iteration order (`storage.rs:2569`), `recall_by_tags` truncates it with no intervening sort (`mod.rs:2629-2631`), and `MemoryId` derives `Ord` (`types.rs:16`), so the sort is arbitrary-but-stable exactly as you describe. `sessions.rs:453-458` already works around this — its comment names the bug outright — so this retires that workaround's rationale.

Two things before merge:

1. **No test covers the change.** Nothing in the repo asserts an order or a set from this path, so "86 passed" doesn't speak to it. Seed more records than the limit under one tag, call `recall_by_tags` twice in one process, assert the same set — then revert the sort and watch it fail. A determinism test can pass on a HashSet by luck, and we have shipped two tests here that could not fail.

2. **"Callers wanting recency already sort again" holds for `sessions.rs` but not for the endpoint this fixes.** `/api/recall/tags` serialises storage order straight to the client (`recall.rs:3474-3480`), so after this PR it returns a deterministic-but-arbitrary 50 of 5000, never the newest 50. The memories are already materialised at `mod.rs:2630`, so a `created_at`-descending sort there costs nothing and fixes determinism and usefulness together. Your call whether that belongs here or in a follow-up — the PR scopes it out honestly either way.

Non-blocking: the `Combined` arm re-collects through a `HashSet` (`storage.rs:2396-2402`), discarding whatever ordering its leaves produced. Latent today — the only live builder (`handlers/search.rs:45-86`) uses entity/date/importance and never tags — but a single `memory_ids.sort()` after the match in `search()` would cover every criterion at once, including this one.

The L1 Smoke Suite failure is main's inherited recall regression (identical lines on three unrelated branches) and is not counted against this PR.

Portll:
Rebased on current main and addressed both, plus the non-blocking one:

1. `recall_by_tags_page_is_repeatable_and_newest_first` (tests/memory_persistence_tests.rs): 40 remembered under one tag, a page of 10 requested twice, same set asserted, and that it is the newest 10 of the 40. Verified failing with both sorts reverted to main: it fails on the same-set assertion.
2. `recall_by_tags` sorts `created_at` descending before it truncates (stable sort over the id-ordered storage result), so `/api/recall/tags` returns the newest `limit`. Done here rather than in a follow-up, since the memories were already materialised at that point.
3. The `Combined` arm sorts after intersecting. I kept it to that arm rather than one `memory_ids.sort()` after the match: `search_by_date` is a keyed prefix scan in date order and `recall_by_date` truncates on it, so a blanket id sort there would turn oldest-first into arbitrary.

Marked ready for review.

Portll:
The Build job failure on this run is `handlers::sources::tests::a_lease_left_by_a_dead_process_is_swept_into_an_aborted_run` panicking on `src/handlers/sources.rs:928` with RocksDB `lock hold by current process … No locks available` on the shared DB LOCK file. This PR does not touch `handlers/sources.rs`; the same job passed on #546 and #551 in the same hour, and 1,29x other tests passed. Looks like a test-isolation flake in that lease test rather than anything here; a re-run should clear it, which I cannot trigger from a fork.

PR #551, opened 2026-09-13
fix(recall-harness): refuse to diff reports from different or unknown embedders

Portll:
## What
`compare_to_baseline` returns an infrastructure failure before comparing any metric when the two reports name different embedders, or when either is empty or the new `EMBEDDER_UNKNOWN`.

## Why
`Report` carries `embedder` and both committed baselines say `minilm-l6-v2`, but nothing acted on the field: `recall.yml` reads it into `BASELINE_EMBEDDER` and `recall_diff.py` renders it in the PR comment, and both only display it. A baseline recorded under a different embedder, or hand-edited, diffed its numbers against the current report and reported the difference as a regression or an improvement, with the mismatch sitting in both files. Two unknowns are not the same embedder; they are no information.

What this closes is the mismatched-or-hand-edited-baseline case. **The runtime case stays open.** `EMBEDDER_ID` is a compile-time constant (`runner.rs`) written into every report, while `SHODH_EMBEDDER` selects the embedder at runtime (`minilm.rs`). A run under `SHODH_EMBEDDER=nomic` therefore produces a report labelled `minilm-l6-v2`, diffs against the MiniLM baseline, and passes this guard cleanly.

## Not in this PR
How the field is populated. Deriving it from the active embedder is the second half of the same defect and a separate change.

## Verification
Two tests: different embedders with identical numbers → one refusal and no metric failure; unknown or empty on either side, including `""` against a real name in both directions → a refusal that names provenance. `cargo test --lib recall_harness::report::tests` and `cargo fmt --all -- --check` on the rebased branch: 23 passed, 0 failed; fmt clean; clippy at CI flags exits 0.

Found by commitwork.


varun29ankuS:
Sound mechanism, and the tests are genuinely mutation-sensitive: the deliberately-failing metrics are what make "before any metric" load-bearing, and `.contains("provenance")` is the assertion that catches removing the unknown check rather than the differs check. No CI risk either — both committed baselines carry `"embedder": "minilm-l6-v2"`, and `embedder` has no `#[serde(default)]`, so a baseline missing the field is already a parse error rather than a silent empty string.

Three fixes:

**1. The const landed inside `Failure`'s doc comment.** `report.rs:435-442`: there is no blank line after the `kind = "case"` line, so lines 435-441 are one contiguous doc block that now attaches to `EMBEDDER_UNKNOWN`, and `Failure` is left undocumented while the const's rustdoc opens with "A failure entry in the report." This is the same defect as item 1 on #546 last week — a new item inserted into the gap between an existing doc block and its owner. Move the const beside `SMOKE_K` (`report.rs:16-17`), where the module's top-level constants live; just adding a blank line would fix `Failure` but leave the const adrift in the failure-type section.

**2. `cargo fmt --all`.** I ran rustfmt against this branch's head: four diffs, at `report.rs:586`, `1082`, `1089`, `1096`. Two of them are already under 100 characters — they exceed `fn_call_width`, not `max_width` — so shortening the long tuple array alone will not clear Format Check. That check gates CI Summary, which is the only required check on main, so it blocks the merge on its own.

**3. The "Why" overclaims, and the open half is the more dangerous one.** A cross-embedder run is still undetected after this change: `EMBEDDER_ID` is a compile-time constant (`runner.rs:38`) written into every report (`runner.rs:414`), while `SHODH_EMBEDDER` selects the embedder at runtime (`minilm.rs:325-345`, which switches nomic to 768 dimensions). A run with `SHODH_EMBEDDER=nomic` therefore produces a report labelled `minilm-l6-v2`, diffs against the MiniLM baseline, and passes the new guard cleanly. What this PR closes is the mismatched-or-hand-edited-baseline case. Worth rewording so the claim matches what the code does, and naming the mislabelling path as still open — your "Not in this PR" note is nearly that, but the "Why" reads as though the runtime case is covered.

Minor precision: "nothing read the field" is very slightly overstated — `.github/workflows/recall.yml:158` reads it into `BASELINE_EMBEDDER` and `scripts/recall_diff.py:136-138` renders it in the PR comment. Both display it; neither acts on it, so the substance stands.

Optional: the loop covers empty only as `("", "")`. Adding `("", "minilm-l6-v2")` and the reverse would match the test's name and pin the `is_empty` branch against the differs branch, which is the discrimination the asymmetric case buys.

The L1 Smoke Suite failure is main's inherited recall regression (exit 1, `EXIT_REGRESSION`; the new guard would exit 2) and is not counted against this PR.

Portll:
Rebased on current main and addressed all three, plus the optional one:

1. `EMBEDDER_UNKNOWN` now sits beside `SMOKE_K` with the module's other top-level constants; `Failure` has its doc block back.
2. `cargo fmt --all`: the four diffs. `--check` is clean on the branch.
3. The description's *Why* is reworded to what the code does: this closes the mismatched-or-hand-edited-baseline case, and the runtime `SHODH_EMBEDDER` mislabelling path is named as still open rather than implied covered. Also corrected "nothing read the field" to "nothing acted on it", citing the two display sites.
4. The loop now covers `("", real)` and `(real, "")` in both directions, pinning the `is_empty` branch apart from the differs branch.

`cargo test --lib recall_harness::report::tests`: 23 passed. Clippy at CI's flags exits 0. Marked ready for review.

PR #555, opened 2026-09-15
feat(encryption): crypto core — keystore, rotation, at-rest encryption of the Memory record (part 1 of #318)

Portll:
## What
Part (1) of the split proposed on #318: the crypto core only — keystore, DEK rotation, and at-rest encryption of the primary `Memory` record. Opt-in and default-off: with no `keystore.json` and no passphrase, nothing here runs.

The bytes on disk are byte-identical to main (pinned by a test). 

Ported hunk by hunk onto current main (`a724b86`) rather than rebased, since `storage.rs` had ~1,000 lines of unrelated change since #318's merge-base. Three commits (all build green for me)

Not in this PR, per the split: KMS integration (part 2), the oblivious seam and PIR (part 3), the HTTP embedder (its own PR), and index blinding (deferred; index keys stay plaintext and `SECURITY.md` says so).

## The blocker from #318, fixed with a contract test
`persist_access_updates` on current main encoded with `encode_sho` directly, so every recalled memory would have had its ciphertext overwritten with plaintext. It and every other memory-record write on main (`store_inner`, `migrate_memory_format`, both forget paths, `migrate_legacy`, `store_with_vectors`, `migration.rs`) now go through one `encode_memory`. `tests/encryption_round_trip.rs` is the test you asked for: store → `get` → access update → drop → raw RocksDB read is an encrypted record with no plaintext substring → reopen and read back.

## Your other before-merge items
- **Encryption requested but unavailable fails loud** — keystore present, passphrase unset or empty: `MemoryStorage::new` is `Err` naming `keystore.json` and `SHODH_MASTER_PASSPHRASE`, and the raw bytes are untouched. `tests/encryption_unavailable_fails_loud.rs`; wrong secret in `tests/encryption_wrong_passphrase.rs`.
- **Tripwire** — a plaintext record read under an active keystore is counted (`plaintext_reads_under_keystore()`), logged at WARN, and lazily re-encrypted on the next write of that record; `SHODH_REQUIRE_ENCRYPTED_READS=1` turns the read into an error. `tests/encryption_plaintext_tripwire_warn.rs` (WARN captured through a subscriber) and `tests/encryption_plaintext_tripwire_strict.rs`.
- **Scope stated honestly** — `SECURITY.md` and `docs/encryption-v2-design.md` are rewritten to what ships: covered = the primary `Memory` record; not covered = facts, graph, vector-index embeddings, the secondary index CF, oplog and sibling CFs, memory-resident plaintext. The PR's earlier claims about HMAC-blinded index terms and KMS are gone.
- **Ops** — `shodh-keyctl` takes secrets from env only (`SHODH_MASTER_PASSPHRASE`, `SHODH_NEW_MASTER_PASSPHRASE`, `SHODH_RECOVERY_CODE`), nothing on argv; keystore files are created `0600` (asserted in a test) with the `.bak` treated the same; `rotate_dek` persists the new keystore before it is used (`tests/encryption_rotation.rs` covers rotate → persist → reopen, plus the rollback-generation guard). Windows ACLs are not done and are documented as such.
- **Rebased on current main** — three commits on `a724b86`, no merge commits.

## Two things to look at
- The process-global crypto is a `RwLock<Option<Arc<StorageCrypto>>>` rather than #318's `OnceLock`, because a `OnceLock` could never pick up a rotated keystore in the same process: post-rotation writes would have kept using epoch 0.
- Tests use the production KDF cost (256 MiB, t=3) with no weakening seam. To keep that runnable under `cargo test`, `Cargo.toml` sets `opt-level = 3` for `argon2` and `blake2` in the dev and test profiles. If you would rather resolve that differently it is one isolated block.

`Cargo.lock` adds seven crates and moves nothing else; nine unrelated edges cargo wanted to re-pick were put back to main's choices and checked with `cargo metadata --locked`.

## Verification
Run twice on the branch, once by the porting session and once independently, plus a control on pristine `a724b86`, all on macOS with Rust 1.97.1:

- `cargo fmt --all -- --check`: clean. Clippy at CI's flags: exit 0, no new warnings in ported files.
- Each commit builds on its own (`cargo check --all-targets` on the keystore-only commit).
- Keystore unit tests 25/25; the nine new integration tests 9/9.
- `cargo test --lib --no-fail-fast`: 1321 passed, 1 failed. The failure is the `handlers::sources` RocksDB LOCK flake already seen on #549/#550; it passed in the full run.
- Full `cargo test --no-fail-fast`: 57 binaries, 2220 passed, 4 failed, 21 ignored. The four are not this branch's: `memory_origin`'s `every_origin_has_a_unique_wire_name_that_round_trips` pins `MemoryOrigin::ALL` at 15 where main has 16, and both `tests/memory_origin.rs` and `src/memory/types.rs` are byte-identical to main; the other three are `timing_sla_tests` debug-profile thresholds. Two (`test_sla_record_p50_latency`, `test_sla_throughput_sustained`) fail on pristine main on the same machine. The third, `test_sla_concurrent_access_latency`, measured 222 ms against its 200 ms threshold in the full run, which shared the machine with three other cargo suites; run alone on the branch afterwards it passes (7.4 s, threshold met), and it passed on pristine main in the control. 

Encryption is off in all three (no keystore), so they do not exercise this code either way.

Found by commitwork.


Portll:
Added `0e159ba`, the same three-line rustls 0.23.43 → 0.23.45 lockfile bump as #554 (cherry-picked with `-x`), so Security Audit and therefore CI Summary can go green here without waiting on #554. Whichever lands first, the other rebases to an identical lock line and the commit drops out. `cargo audit` on this branch: 0 vulnerabilities, the same 5 allowed warnings; `cargo metadata --locked` accepts the lock.

The remaining red check, L1 Smoke Suite, is main's inherited recall regression: the same two lines (`p@1` 0.31 → 0.30, `p@1[multi_hop]` 0.28 → 0.24) appear on #554, which changes nothing but three lockfile lines, and on #546. It is in the Recall Harness workflow, not in CI Summary's `needs`.

varun29ankuS:
Thanks for this, Portll. The port is careful work, and the #318 blocker is really fixed. I went through every path on this branch that writes a Memory record: store, update/modify, access updates, both forget paths, lazy and bulk migration, `store_with_multimodal_vectors`, `migration.rs`, and the retry buffer. All of them go through `encode_memory`, and I found no bypass. `encryption_default_off.rs` is a good test: it compares raw bytes after an access bump and checks that the bump changed them, so it can actually fail. All nine encryption tests pass for me on Windows.

It isn't ready to merge yet. What's left is one design gap that shows up three ways: **the database doesn't remember which keystore encrypted it.** The generation sentinel protects against rollback, but it doesn't tie a keystore to a database. Each of these was reproduced with a small test:

1. **Lost keystore → silent recreate → cleanup deletes data.** Store a record, delete `keystore.json`, restart with the passphrase set. `init_storage_crypto` takes the `(false, Some(pass))` branch and creates a fresh keystore without checking whether the DB already holds `ENC\0` records. A gen-0 DB has no sentinel, so the rollback check doesn't catch it. `get` then fails with `aead::Error`, and `POST /api/storage/cleanup` deletes the record, because `can_decrypt` is true (epoch 0 exists, just with the wrong DEK). The repro ended with `CLEANUP_DELETED=1`. AEAD can't tell a wrong key from tampered bytes, so the "authenticated corruption" branch in `cleanup_corrupted` shouldn't delete an `ENC\0` record under any circumstances. This also contradicts the PR description's "never deleted by the corruption cleanup". Since `/api/backup` doesn't include `keystore.json`, restoring a backup onto a new host hits this path.
2. **HTTP multi-tenant.** The keystore lives at `storage_path.join("keystore.json")`, and each user's storage path is their own directory (`user_dir(user_id)` in `state.rs`). So each user gets their own keystore with a random KEK, and the process-global `STORAGE_CRYPTO` slot refuses every user after the first ("a different encryption keystore is already active in this process"). With `SHODH_MASTER_PASSPHRASE` set, only whichever user loads first works, and the refused user's `keystore.json` has already been written. `encryption_keystore_guard.rs` asserts that refusal as correct, but in server mode it takes every other tenant offline. MCP's single default user hides this.
3. **Accepted rollback keeps the stale cryptor.** Rotate to gen 1 and reopen. Then restore the `.bak` with `SHODH_ALLOW_KEYSTORE_ROLLBACK=true` and reopen in the same process. The `existing.generation > fresh.generation => return Ok(())` branch keeps the newer cryptors, so new writes go out under epoch 1 while the keystore on disk only has epoch 0 (`RECORD_EPOCH=Some(1)`, on-disk epochs `[0]`). After a restart those records are unreadable.

The invariant I think is missing: *a database that has ever been encrypted must remember its keystore (a fingerprint, plus "has been encrypted"), and open must refuse to create or accept a different one.* I'll leave the design to you. A keystore at the data root shared across users, per-store cryptors instead of a process-global slot, and a fingerprint stored in the DB are all possible, and each changes the multi-tenant answer differently.

**Scope docs.** The tantivy BM25 index stores full content (`hybrid_search.rs`, `content` is `TEXT | STORED`, under `<user>/bm25_index`), and neither SECURITY.md nor the design doc lists it. Under the cold-disk threat model that is a complete plaintext copy of every memory. The audit log's `content_preview` also belongs on the not-covered list, along with a note that the plaintext vector-index embeddings can be inverted back to content. I don't expect part 1 to encrypt these. The docs just need to say so plainly.

**Punch-list items still open or partial:**
- The KDF floor is 8 MiB (`keystore.rs:52`). We asked for at least 19 MiB.
- DEK wraps still use the raw KEK. The MAC key is derived; the wrap key isn't.
- `recovery_key_from_code` isn't zeroized.
- The sentinel isn't authenticated, and a wrong-length sentinel is read as 0, which turns the guard off. I'd make that an error.
- SECURITY.md should carry the Windows ACL caveat and the "back up `keystore.json` separately" note.

Smaller items:
- `get_opt`'s lazy re-encrypt runs outside `record_mutation_lock`, so it can overwrite a concurrent `modify`. The race predates this PR, but it now fires on every plaintext read.
- Enabling encryption over an existing store leaves old plaintext in SSTs and backups until compaction. Worth a line in the docs.

The `RwLock` over `OnceLock` choice is right, and so is keeping the production KDF cost in tests. Once the keystore↔DB binding is in, with tests for these three scenarios, I'm happy to look again.

